PR #1, opened 2021-12-16
(#1) basic test framework and mock service components

eli-darkly:
This PR creates the overall project layout and provides infrastructure that will be used for SDK tests, even though there aren't yet any tests so it doesn't do anything.

Everything in the `framework` package and its subpackages is based on the corresponding package in https://github.com/launchdarkly/sse-contract-tests, although I've reworked it considerably. That's the part that isn't domain-specific to SDK testing— it's just the low-level functionality for communicating with a test service, and the basic test runner API that deliberately resembles Go's `testing` package. I'd like to eventually have the `sse-contract-tests` project share that same code directly (either in its own repo, or by importing it from the `sdk-contract-tests` codebase).

The `mockld` package provides two of the test fixtures we'll be needing: a fake streaming service and a fake event-recorder. Eventually we'll want to add a fake polling service, and add some APIs to all of these for simulating error conditions. But this gives us the basic ability to feed data into an SDK and receive events.

The `servicedef` package contains the Go definition of the test service protocol. This is explained in more detail in `docs/service_def.md`.

The `sdktests` package is where the high-level test logic will be. Currently the test suite (`testsuite_server_side.go`) is empty, but the package also contains domain-specific APIs to be used by tests.


keelerm84 (approved):

main.go
```diff
@@ -0,0 +1,83 @@
...
+	ldtest.PrintResults(results)
+
+	if params.stopServiceAtEnd {
+		fmt.Println("Stopping test service")
+		if err := harness.StopService(); err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to stop test service: %s\n", err)
+		}
+	}
+	if !results.OK() {
+		os.Exit(1)
```
cwaldren-ld, line 81: It might be useful to distinguish between "the service had a generic failure" and "the service succeeded, but the tests failed." 

Might we return a different error code for this, or let the user specify it like `--exit-error=10` or something similar?

mockld/event_data.go
```diff
@@ -0,0 +1,148 @@
...
+
+func ExpectedEventUserFromUser(user lduser.User, eventsConfig servicedef.SDKConfigEventParams) EventUser {
+	// This simulates the expected behavior of SDK event processors with regard to redacting
+	// private attributes.
+	allJSON := helpers.AsJSONValue(user)
+	o := ldvalue.ObjectBuild()
+	var custom ldvalue.ObjectBuilder
+	var private []ldvalue.Value
+	allAttributes := append(allJSON.Keys(), allJSON.GetByKey("custom").Keys()...)
+	for _, attr := range allAttributes {
```
cwaldren-ld, line 89: I'm having some trouble understanding the intent of this loop. Would there be a way to split it into multiple passes, or otherwise refactor out some of the tests into named methods?

mockld/events_service.go
```diff
@@ -0,0 +1,79 @@
...
+type EventsService struct {
+	AnalyticsEventPayloads chan Events
+	sdkKind                SDKKind
+	credential             string
+	logger                 framework.Logger
+}
+
+func NewEventsService(sdkKind SDKKind, credential string, logger framework.Logger) *EventsService {
+	return &EventsService{
+		AnalyticsEventPayloads: make(chan Events, 100),
```
cwaldren-ld, line 23: Is 100 arbitrary?

mockld/sdk_data.go
```diff
@@ -0,0 +1,147 @@
...
+	ServerSideSDK SDKKind = "server"
+	MobileSDK     SDKKind = "mobile"
+	JSClientSDK   SDKKind = "jsclient"
+)
+
+type DataItemKind string
+
+type SDKData interface {
+	SDKKind() SDKKind
+	Serialize() []byte
```
cwaldren-ld, line 25: Will this serialization always be JSON?

framework/harness/harness.go
```diff
@@ -0,0 +1,138 @@
...
+	deadline := time.NewTimer(httpListenerTimeout)
+	defer deadline.Stop()
+	ticker := time.NewTicker(time.Millisecond * 10)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-deadline.C:
+			return fmt.Errorf("could not detect own listener at %s", server.Addr)
+		case <-ticker.C:
+			resp, err := http.DefaultClient.Head(fmt.Sprintf("http://localhost:%d", port))
```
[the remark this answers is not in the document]
eli-darkly: "It may not make progress for a couple iterations" is not a problem; we're not concerned about wasting a few CPU cycles during the startup of this process.

The goal here is not to make the minimum possible number of checks, but to detect success as soon as possible so we don't delay for an unnecessarily long time. If you changed the `time.NewTicker` parameter to half a second, what would happen is that the first listener check would fail (very likely, since `NewTicker` sends a tick _immediately_) and then the app would wait half a second. Every startup, or nearly every startup, would take at least half a second.

framework/helpers/json.go
```diff
@@ -0,0 +1,45 @@
...
+	"encoding/json"
+	"sort"
+	"strings"
+
+	"gopkg.in/launchdarkly/go-sdk-common.v2/ldvalue"
+)
+
+// AsJSON is just a shortcut for calling json.Marshal and taking only the first result.
+func AsJSON(value interface{}) []byte {
+	ret, _ := json.Marshal(value)
```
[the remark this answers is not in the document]
eli-darkly: Yes.

framework/ldtest/test_scope.go
```diff
@@ -0,0 +1,185 @@
...
+)
+
+type environment struct {
+	config  TestConfiguration
+	results Results
+}
+
+// T represents a test scope. It is very similar to Go's testing.T type.
+type T struct {
+	env         *environment
```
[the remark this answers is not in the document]
eli-darkly: It's an object that's shared by all test scopes. We could copy it by value, it does not get modified, but if it _did_ get modified we would not want the change to affect only a single T.

main.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/launchdarkly/sdk-test-harness/framework/ldtest"
+	"github.com/launchdarkly/sdk-test-harness/sdktests"
+	"github.com/launchdarkly/sdk-test-harness/servicedef"
+)
+
+const defaultPort = 8111
+const statusQueryTimeout = time.Second * 10
+
+//go:embed VERSION
+var versionString string // comes from the VERSION file which we update for each release
```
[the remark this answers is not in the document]
eli-darkly: That's an interesting feature but I'm not sure how reliable the module version would be in this context. The application binary is (I think) getting built before a new version tag has been created, so it would probably be picking up the _previous_ module version.

main.go
eli-darkly: I'm not sure what "the service had a generic failure" means. If the test harness was unable to run, because its logic is broken somehow or because the test service didn't accept its requests, then we have to treat that as a failure of the SDK contract tests. Either case requires human intervention to look into the situation and see what needs to be fixed, starting with looking at the errors that were logged. There's no way to save ourselves any time in that regard by returning a different exit code programmatically.

mockld/event_data.go
eli-darkly: Well, this is imitating code that we already implemented in https://github.com/launchdarkly/go-sdk-events which is the standard behavior of all of the LD SDKs regarding private attributes in events— as the comment says. And the behavior of private attributes is documented in every SDK, so I'd rather not repeat the explanation in detail here, and I don't want to make the logic extra long by adding unnecessary multiple passes. I can add a couple more comments though.

mockld/events_service.go
eli-darkly: Yes.

mockld/sdk_data.go
eli-darkly: Yes. That is how all of the LD service endpoints work, as documented in the specs for the SDK data model and the streaming and polling services.

framework/helpers/json.go
```diff
@@ -0,0 +1,45 @@
+package helpers
```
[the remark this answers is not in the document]
eli-darkly: Yes, it's all going away, sorry I didn't make that clearer in this PR.

mockld/events_service.go
eli-darkly: The intention is just to make it unlikely for us to get blocked by a channel being full. The goroutine that listens for event posts should be able to just dump the data onto the channel and call it a day; there's no value in this case in having it block till we've read the value from the channel.

framework/helpers/json.go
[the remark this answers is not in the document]
eli-darkly: That is... in this case it is safe, and in general it is almost always safe to return nil for a `[]byte` result.

The reason is that there are only a limited number of things you can do with a byte slice:

1. You could get its length with `len`.
2. You could convert it to a string.
3. You could try to get an individual byte element from it.
4. You could get a sub-slice of it

1 and 2 are literally _always_ safe; converting `nil` to a string just gives you `""`, and using `len` gives you zero.

3 and 4 are potentially unsafe with nil— but only in the same sense that they're potentially unsafe with a non-nil slice too, because they can panic if you use an out-of-range index. So you would always want to do a length check first, in which case again `nil` behaves just the same as a zero-length slice.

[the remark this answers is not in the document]
eli-darkly: So, even though I said "almost always", I'm actually having trouble thinking of any case where it is not safe. I guess maybe if you're doing direct pointer manipulation stuff with the `unsafe` package, there might be some situation where treating a nil slice value as a pointer could get you in trouble. So there might be an unsafe situation if you're using the package that is actually named that.

mockld/event_data.go
eli-darkly: Done - hope that's clearer.

mockld/events_service.go
cwaldren-ld: I think it'd be worth encoding this answer into a comment on a constant written in the code.

cwaldren-ld (approved):

PR #3, opened 2021-12-16
(#3) implement basic analytics event tests

eli-darkly:
This PR adds some tests for analytics event behavior, roughly equivalent to unit tests that most of the server-side SDKs have:

* Evaluations generate `feature` events when appropriate, with the expected properties, and `summary` events
* `index` events are also generated as appropriate based on the SDK configuration
* `identify`, `custom`, and `alias` events are generated by the corresponding SDK methods
* Private attributes behave correctly when users appear in any of these events

It does _not_ include:

* Any tests of the details of summary event counters
* Verifying retry behavior for failed HTTP posts
* Verifying that the SDK `capacity` parameter (event buffer size) can correctly drop excess events

Once those things are added, this will be at least as good as the unit test coverage for event behavior that we've repeatedly implemented in the SDK projects. It is very similar to the integration-harness coverage, but better in a couple of ways:

* Testing SDK configuration properties like `allAttributesPrivate` and `privateAttributeNames`, which are set at SDK initialization time rather than per user, is easy since we can create a fresh SDK client instance for each test. In integration-harness, this requires maintaining an extra SDK client in each restwrapper instance that has `allAttributesPrivate` set; we didn't bother testing the global `privateAttributeNames` in integration-harness at all.
* The coverage for custom event parameters is better, since it's easy to run a parameterized test with all possible combinations of "is there a data parameter", "what type is it", and "is there a metricValue parameter". In integration-harness we would need to write out a fairly verbose piece of test data for each.
* Feedback for test failures is much better, because we can see exactly what events were generated by each test right away instead of having to wait and sift through a huge flood of all the events from all tests.
* Also, we don't need to do nearly as much testing for feature events. Integration-harness doesn't really have a separate test suite for this— instead, it creates an event expectation for every single evaluation test. But the ability of the SDK to copy evaluation parameters and results into an event is orthogonal to what kind of evaluation it is, so we only need to run test cases for the combinations of parameters that really matter.

If you'd like to see this running against a basically functional implementation of the test service for the Java SDK:

1. Check out the `sdk-test-harness` branch of `java-server-sdk-private`.
2. In that directory, run: `make build-contract-tests start-contract-test-service`
3. In another terminal, check out this PR branch of `sdk-test-harness` and run `make`; then run: `./sdk-test-harness -url http://localhost:8000`

You'll notice that the Java SDK does fail one test ("events/identify events/user with empty key generates no event"). That's because its behavior is inconsistent with the behavior of the Go SDK in that regard, which I originally wrote the test against. Whether that's a legit bug, or an acceptable variation (meaning I should remove the test), I'm not sure. You'll have to scroll up to see the error details because the logging is not yet very convenient.


sdktests/server_side_events_custom.go
```diff
@@ -0,0 +1,128 @@
...
+	dataSource := NewSDKDataSource(t, mockld.EmptyServerSDKData())
+	events := NewSDKEventSink(t)
+	client := NewSDKClient(t, WithConfig(servicedef.SDKConfigParams{Events: &eventsConfig}), dataSource, events)
+
+	user := lduser.NewUser("user-key")
+
+	// Generate many permutations of 1. data types that can be used for the data parameter, if any, and
+	// 2. metric value parameter, if any.
+	allParams := make([]servicedef.CustomEventParams, 0)
+	omitMetricValue := float64(-999999) // magic value that we'll change to null
```
cwaldren-ld, line 69: I wonder if https://pkg.go.dev/testing/quick could be utilized at all here? Perhaps it doesn't fit the use-case.

cwaldren-ld (approved):

sdktests/server_side_events_custom.go
eli-darkly: Well... I see what you're getting at, but I'm not really trying to do black-box testing here in that sense. Maybe it's worth talking (outside of this PR) about whether we should do more in that regard, but in the current overall design of these tests I did make a firm decision not to go that route.

Pure black-box testing, when numeric or string parameters are involved, would imply running millions or billions of permutations for basic operations even when there's no _functional_ difference between the values. That is, in an operation where the value isn't used in any computation that could ever produce an error for value A as opposed to value B—for instance, if the value is simply being copied unchanged into the output—in pure black-box testing you're not allowed to know any such implementation details, so you still need to do just as many tests.

But if you're allowed to make at least very basic assumptions about what kinds of differences in behavior for different inputs might be plausible, and about what layer of the software you are trying to test, and what kinds of variances are or are not meaningful, then a more focused approach is possible. The assumptions being made in this case with custom events are:

* The `data` and `metricValue` parameters, if present, should cause those same values to be copied into the output event. They are not interpreted by SDKs in any other way.
* Event data is JSON data. The `data` parameter can be any JSON value. JSON is already a well-defined spec and if the overall implementation of JSON being used by the SDK was broken, that is a bigger problem that we'd expect to detect in tests that aren't specific to custom events. So in this test we aren't concerned with, for instance, a hypothetical error where the string "abc" gets encoded correctly but the string "abcd" does not.
* However, on most platforms, JSON data has to be converted to and from some built-in data types. So it is possible that an SDK could mistakenly declare `data` as a string, in which case it would work correctly for all string values but fail if `data` was a JSON number. Therefore it makes sense to test separate cases where `data` is each of the allowable JSON types.
* Similarly, the semantics of `metricValue` are that it is either a number (any number) or "no value". It's certainly possible that an SDK could handle one of those cases correctly but not the other, so we should test both.
* It is theoretically possible that an SDK could work for one numeric value of `metricValue` but not another, even though (in terms of how this method is defined) there's no reason for there to be any computations with it at all. But if our test is "the same number comes out that we put in", we can't blindly test it with every possible `float64` value, because JSON itself couldn't pass such a test— many `float64` numbers can't be represented accurately in a decimal format. That's a well-known limitation of any system that uses JSON numbers, and LaunchDarkly simply can't guarantee that kind of accuracy.

So, my compromise on that last point was to use just a couple of example numeric values which are meant to catch possible misuse of strong typing in the SDK logic. That is, even if the SDK doesn't do any computations with the value, if it mistakenly used an integer type or an unsigned type then it would mangle non-integer or negative values. Or if it assumed that zero meant "no value" (i.e. it didn't use a nullable type) then you wouldn't be able to send a zero value, so there's a test case for that.

PR #6, opened 2022-01-20
add parameterized tests for date & semver operators

eli-darkly:
There's no new Go logic for these - I'm just using the existing file-driven parameterized test.

I used slightly different strategies in each file for how to iterate over all the desired values - see PR comments. It might be possible in the future to redesign the syntax & the parser to make these clearer.

### DIFF
--- a/testdata/data-files/server-side-eval/operators-date-equal-or-invalid.yml
+++ b/testdata/data-files/server-side-eval/operators-date-equal-or-invalid.yml
@@ -1,0 +1,89 @@
+---
+name: operators - date - equal or invalid
+
+constants:
+  IS_MATCH:
+    value: true
+    variationIndex: 0
+    reason: { kind: "RULE_MATCH", ruleIndex: 0, ruleId: "ruleid" }
+  IS_NOT_MATCH:
+    value: false
+    variationIndex: 1
+    reason: { kind: "FALLTHROUGH" }
+  base_flag: &base_flag
+    on: true
+    fallthrough: { variation: 1 }
+    variations: [ true, false ]
+  base_rule: &base_rule
+    id: ruleid
+    variation: 0
+
+parameters:
+  # do permutations of each VALUE1 with each VALUE2; these are values that are either invalid as dates,
+  # or valid as dates but equal to each other
+  -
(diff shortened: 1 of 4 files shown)


testdata/data-files/server-side-eval/operators-date-equal-or-invalid.yml
```diff
@@ -0,0 +1,89 @@
...
+    on: true
+    fallthrough: { variation: 1 }
+    variations: [ true, false ]
+  base_rule: &base_rule
+    id: ruleid
+    variation: 0
+
+parameters:
+  # do permutations of each VALUE1 with each VALUE2; these are values that are either invalid as dates,
+  # or valid as dates but equal to each other
```
eli-darkly, line 23: This YAML is the equivalent of writing the following in JSON:
```json
[
  [
    {
      "VALUE1": "not valid"
    },
    {
      "VALUE1": true
    }, ...etc.
  ],
  [
    {
      "VALUE2": "not valid"
    }, ...etc.
  ]
]
```

The effect, as described [here](https://github.com/launchdarkly/sdk-test-harness/blob/master/docs/data_files.md), is to apply the whole file with {VALUE1: "not valid", VALUE2: "not valid"}, and then with {VALUE1: true, VALUE2: "not valid"}, etc., through every permutation.

testdata/data-files/server-side-eval/operators-date-unequal.yml
```diff
@@ -0,0 +1,83 @@
...
+constants:
+  IS_MATCH:
+    value: true
+    variationIndex: 0
+    reason: { kind: "RULE_MATCH", ruleIndex: 0, ruleId: "ruleid" }
+  IS_NOT_MATCH:
+    value: false
+    variationIndex: 1
+    reason: { kind: "FALLTHROUGH" }
+  base_flag: &base_flag
```
eli-darkly, line 13: This `&` syntax, and the `<<: *` references below, isn't a feature of the sdk-test-harness parser; it's just a mixin/macro syntax that's built into YAML.

keelerm84 (approved):

PR #8, opened 2022-01-21
improve event tests, use new matchers, add summary tests

eli-darkly:
* New tests: detailed expectations about summary event counters
* New tests: evaluating flags with prerequisites causes the prereq evaluations to show up in feature events and/or summary counters
* New test: if a flag evaluation fails (like with a malformed flag error) it can still generate feature/summary events
* New tests for enforcement of event buffer capacity— this is a kind of test that was not possible in integration-harness (**Note:** this requires the test service to implement a new config option, `events.capacity`)
* Miscellaneous updates throughout the event tests to take advantage of the newer `matchers` API in `go-test-helpers`— this is currently pointing to a development branch, pending https://github.com/launchdarkly/go-test-helpers/pull/12


mockld/streaming_service_test.go
```diff
@@ -40,7 +40,7 @@ func TestStreamingServiceServerSide(t *testing.T) {
 
 		initialEvent := requireEvent(t, stream)
 		assert.Equal(t, "put", initialEvent.Event())
-		m.AssertThat(t, initialEvent.Data(), m.JSONStrEqual(expectedPutData(initialData)))
+		m.In(t).Assert(initialEvent.Data(), m.JSONStrEqual(expectedPutData(initialData)))
```
eli-darkly, line 43: This was just an API change in `go-test-helpers`.

sdktests/server_side_eval.go
```diff
@@ -78,31 +78,44 @@ func RunParameterizedServerSideEvalTests(t *ldtest.T) {
...
+	flagKey := "some-flag"
+	user := lduser.NewUser("user-key")
+	expectedReason := ldreason.NewEvalReasonError(ldreason.EvalErrorClientNotReady)
+
 	dataSource := NewSDKDataSource(t, mockld.BlockingUnavailableSDKData(mockld.ServerSideSDK))
 	client := NewSDKClient(t,
 		WithConfig(servicedef.SDKConfigParams{StartWaitTimeMS: 1, TimeoutOK: true}),
 		dataSource)
 
-	for _, suite := range getAllServerSideEvalTestSuites(t, "server-side-eval-client-not-ready") {
```
eli-darkly, line 86: There wasn't really a good reason to use the parameterized data file system for this test, because the only parameter is the value type, and we already have a fairly convenient way to do that programmatically.

sdktests/server_side_events_alias.go
```diff
@@ -17,13 +17,11 @@ type aliasEventTestScenario struct {
 }
 
 func doServerSideAliasEventTests(t *ldtest.T) {
-	eventsConfig := baseEventsConfig()
-
 	userFactory := NewUserFactory("doServerSideAliasEventTests")
 
 	dataSource := NewSDKDataSource(t, mockld.EmptyServerSDKData())
 	events := NewSDKEventSink(t)
-	client := NewSDKClient(t, WithConfig(servicedef.SDKConfigParams{Events: &eventsConfig}), dataSource, events)
```
eli-darkly, line 26: This no longer needs to be specified, unless you're customizing event properties— if you include an `EventSink` then it will assume there should be at least a default event configuration.

sdktests/server_side_events_summary.go
```diff
@@ -0,0 +1,277 @@
...
+	// evaluations for flag2: one for userA producing value2a
+	_ = client.EvaluateFlag(t, servicedef.EvaluateFlagParams{FlagKey: flag2.Key, User: &userA, DefaultValue: default2})
+
+	client.FlushEvents(t)
+	payload := events.ExpectAnalyticsEvents(t, defaultEventTimeout)
+
+	m.In(t).Assert(payload, m.Items(
+		EventIsIndexEvent(mockld.SimpleEventUser(userA)),
+		EventIsIndexEvent(mockld.SimpleEventUser(userB)),
+		m.JSONMap().Should(m.MapOf(
```
eli-darkly, line 61: This kind of nested map assertion is a feature that was added in https://github.com/launchdarkly/go-test-helpers/pull/12. It's different from just doing a deep-equality test on a map, because you can nest other kinds of assertions inside it (like `ItemsInAnyOrder`).

sdktests/custom_matchers.go
```diff
@@ -145,3 +149,22 @@ func EventIsIndexEvent(eventUser mockld.EventUser) m.Matcher {
...
+
+func EventIsSummaryEvent() m.Matcher {
+	return EventHasKind("summary")
+}
+
+func ValueIsPositiveNonZeroInteger() m.Matcher {
+	return m.New(
+		func(value interface{}) bool {
+			v := ldvalue.Parse(jsonhelpers.ToJSON(value))
+			return v.IsNumber() && v.IsInt() && v.IntValue() > 0
```
cwaldren-ld, line 161: Is `IsNumber` redundant?

sdktests/server_side_events_buffer.go
```diff
@@ -0,0 +1,91 @@
...
+
+func doServerSideEventBufferTests(t *ldtest.T) {
+	capacity := 20
+	eventsConfig := baseEventsConfig()
+	eventsConfig.Capacity = capacity
+
+	userFactory := NewUserFactory("doServerSideEventCapacityTests",
+		func(b lduser.UserBuilder) { b.Name("my favorite user") })
+	users := make([]lduser.User, 0)
+	for i := 0; i < capacity+3; i++ {
```
cwaldren-ld, line 24: I'm assuming the `+3` has to be any number > 0? Could give it a name like `const margin = 3` or something?

cwaldren-ld (approved):

sdktests/custom_matchers.go
eli-darkly: Yeah

PR #9, opened 2022-01-22
add tests for stream behavior, updates, and summary events

eli-darkly:
This PR adds some more big areas of core functionality to the contract tests, including quite a few things that can't be done by integration-harness:

* General stream HTTP and retry behavior (see tasks in [sc-138807](https://app.shortcut.com/launchdarkly/story/138807/tests-for-stream-behavior))
* Updates are stored correctly with versioning when received from the stream ​(see [sc-138804](https://app.shortcut.com/launchdarkly/story/138804/tests-for-streaming-flag-updates))
* Summary event counters work correctly when there's more than one version of a flag (this was not included in the previous summary event tests, because it requires sending a stream update)

I believe sdk-test-harness's test coverage in these areas is now on par with the very thorough unit tests in the Go and Java SDKs, except in one area: it's not possible to _directly_ query user segments via the SDK API, so the tests for updating/deleting segments are done indirectly by evaluating a flag.

There are breaking changes to the test service API here, sorry:

* The name of the top-level config parameter `timeoutOk` was changed to `initCanFail`, since timeouts are not the only kind of initialization error we're testing now (it could also be a 401 error, for instance).
* There's also a new parameter in the streaming config section, `initialRetryDelayMs`.

### DIFF
--- a/mockld/streaming_service.go
+++ b/mockld/streaming_service.go
@@ -24,9 +24,7 @@ func (l eventSourceDebugLogger) Printf(fmt string, args ...interface{}) {
 }
 
 type StreamingService struct {
-	URL         string
 	sdkKind     SDKKind
-	credential  string
 	initialData SDKData
 	streams     *eventsource.Server
 	debugLogger framework.Logger
@@ -35,7 +33,7 @@ type StreamingService struct {
 
 type eventImpl struct {
 	name string
-	data map[string]interface{}
+	data interface{}
 }
 
 const (
@@ -43,7 +41,6 @@ const (
 )
 
 func NewStreamingService(
-	credential string,
 	initialData SDKData,
 	debugLogger framework.Logger,
 ) *StreamingService {
@@ -53,7 +50,6 @@ func NewStreamingService(
 
 	s := &StreamingService{
 		sdkKind:     initialData.SDKKind(),
-		credential:  credential,
 		initialData: initialData,
 		streams:     streams,
 		debugLogger: debugLogger,
@@ -68,11 +64,6 @@ func (s *StreamingService) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	case r.URL.Path == "/all" && s.sdkKind == ServerSideSDK:
 		if r.Method != "GET" {
 			w.WriteHeader(http.StatusMethodNotAllowed)
-			return
-		}
-		authKey := r.Header.Get("Authorization")
-		if authKey != s.credential {
-			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}
 		s.streams.Handler(allDataChannel)(w, r)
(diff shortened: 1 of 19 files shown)


mockld/streaming_service.go
```diff
@@ -24,9 +24,7 @@ func (l eventSourceDebugLogger) Printf(fmt string, args ...interface{}) {
 }
 
 type StreamingService struct {
-	URL         string
 	sdkKind     SDKKind
-	credential  string
```
eli-darkly, line 29: I removed SDK key validation from this low-level component; the mock endpoints will now accept the request no matter what. Instead, there's now a specific test in `server_side_stream_request.go` that looks at the Authorization header after the client has connected.

mockld/streaming_service.go
```diff
@@ -35,15 +33,14 @@ type StreamingService struct {
 
 type eventImpl struct {
 	name string
-	data map[string]interface{}
+	data interface{}
```
eli-darkly, line 36: Changed how mock stream events are specified so that the data can be any JSON-serializable value or a `json.RawMessage`; that way we can inject more kinds of arbitrary data into the stream.

sdktests/server_side_data_store_updates.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func doServerSideDataStoreStreamUpdateTests(t *ldtest.T) {
+	// These tests verify that the SDK's default data store correctly implements updates and
+	// deletes with versioning.
+	//
+	// Since our mechanism for putting updates into the SDK is our stream test fixture, these
+	// tests also verify that the stream is working correctly.
+
```
eli-darkly, line 27: We have unit tests equivalent to this in most/all of the SDKs, but they're tedious to write and it's easy to miss edge cases.

Integration-harness can only verify the basic "an update is persisted if it has a higher version" case, because integration-harness has no way of causing LD to emit out-of-order updates.

sdktests/server_side_events_summary.go
```diff
@@ -259,6 +263,60 @@ func doServerSideSummaryEventPrerequisitesTest(t *ldtest.T) {
 	))
 }
 
+func doServerSideSummaryEventVersionTest(t *ldtest.T) {
+	// This test verifies that if the version of a flag changes within the timespan of one event payload,
+	// evaluations for each version are tracked separately. We do this by evaluating the flag in its
+	// original version, then pushing a stream update and polling until the SDK reports the updated
+	// value, and then checking that both versions appear in the summary event. More detailed testing of
+	// update behavior is covered in server_side_data_store_updates.go.
```
eli-darkly, line 271: Integration-harness sort of covers this case, but its version of the test is less specific and harder to debug, because the summary events for the updated flag are mixed in with the results of all the other evaluations done during the test run.

sdktests/server_side_stream_request.go
```diff
@@ -0,0 +1,52 @@
...
+			Credential: sdkKey,
+		}), dataSource)
+
+		request := expectRequest(t, dataSource.Endpoint(), time.Second)
+
+		assert.Equal(t, sdkKey, request.Headers.Get("Authorization"))
+		assert.NotEqual(t, sdkKey, request.Headers.Get("User-Agent"))
+	})
+
+	t.Run("URL path is correct when base URI has a trailing slash", func(t *ldtest.T) {
```
eli-darkly, line 29: We've been burned by this before in several SDKs, due to overly simple-minded path concatenation logic.

sdktests/server_side_stream_retry.go
```diff
@@ -0,0 +1,256 @@
...
+const briefDelay ldtime.UnixMillisecondTime = 1
+
+func baseStreamConfig(endpoint *harness.MockEndpoint) *servicedef.SDKConfigStreamingParams {
+	return &servicedef.SDKConfigStreamingParams{
+		BaseURI:             endpoint.BaseURL(),
+		InitialRetryDelayMs: timeValueAsPointer(briefDelay),
+	}
+}
+
+func doServerSideStreamRetryTests(t *ldtest.T) {
```
eli-darkly, line 29: As with the update tests, these are implemented pretty tediously in each SDK's unit tests and we've often forgotten some edge cases there. It's impossible for integration-harness to test any of these scenarios.

sdktests/server_side_stream_validation.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/launchdarkly/go-test-helpers/v2/jsonhelpers"
+	m "github.com/launchdarkly/go-test-helpers/v2/matchers"
+	"github.com/launchdarkly/sdk-test-harness/framework/ldtest"
+	"github.com/launchdarkly/sdk-test-harness/mockld"
+	"github.com/launchdarkly/sdk-test-harness/servicedef"
+	"gopkg.in/launchdarkly/go-sdk-common.v2/lduser"
+	"gopkg.in/launchdarkly/go-sdk-common.v2/ldvalue"
+)
+
+func doServerSideStreamValidationTests(t *ldtest.T) {
```
eli-darkly, line 17: The comment in the previous file also applies here. Integration-harness can't simulate any of these conditions because there's no way to cause a real LD stream to emit these kinds of unexpected data; the SDKs have to be able to handle them in case a stream gets garbled by network problems, gremlins, etc.

keelerm84 (approved):

PR #11, opened 2022-01-24
add ability to write a JUnit XML file + misc logging improvements

eli-darkly:
This makes it so you can add `-junit my_file_name.xml` to the command-line parameters and it will write test result in JUnit XML format to `my_file_name.xml`. This standard format can be parsed by CircleCI so you'll see a nice summary of test failures at the top of your CI job, rather than having to drill down into the sdk-test-harness console output.


docs/running.md
```diff
@@ -3,22 +3,23 @@
 ## Test harness command line
 
 ```shell
-./sdk-test-harness --url <test service base URL> [other options]
+./sdk-test-harness -url <test service base URL> [other options]
```
eli-darkly, line 6: The Go command-line flags parser actually accepts either `-` or `--`.

framework/ldtest/junit_test_logger.go
```diff
@@ -0,0 +1,206 @@
...
+	Failures   int                `xml:"failures,attr"`
+	Time       string             `xml:"time,attr"`
+	Name       string             `xml:"name,attr"`
+	Properties []jUnitXMLProperty `xml:"properties>property,omitempty"`
+	TestCases  []jUnitXMLTestCase `xml:"testcase"`
+}
+
+type jUnitXMLTestCase struct {
+	XMLName     xml.Name             `xml:"testcase"`
+	Classname   string               `xml:"classname,attr"`
```
eli-darkly, line 52: There isn't really a "class" associated with any of these tests, so currently I'm leaving this field blank, which I think CircleCI will be OK with. The `go-junit-report` tool normally sets this to the _package_ name, but there isn't really a package for this either other than "sdktests".

framework/ldtest/junit_test_logger.go
```diff
@@ -0,0 +1,206 @@
...
+	XMLName xml.Name            `xml:"testsuites"`
+	Suites  []jUnitXMLTestSuite `xml:"testsuite"`
+}
+
+type jUnitXMLTestSuite struct {
+	XMLName    xml.Name           `xml:"testsuite"`
+	Tests      int                `xml:"tests,attr"`
+	Failures   int                `xml:"failures,attr"`
+	Time       string             `xml:"time,attr"`
+	Name       string             `xml:"name,attr"`
```
eli-darkly, line 45: I'm setting the test suite name to the string representation of the full test ID, such as `events/summary events/basic counter behavior`.

main.go
```diff
@@ -47,36 +47,54 @@ func main() {
...
-	capabilities := harness.TestServiceInfo().Capabilities
+	enabledCapabilities := harness.TestServiceInfo().Capabilities
+	var allCapabilities framework.Capabilities
 	switch {
-	case capabilities.Has(servicedef.CapabilityServerSide):
+	case enabledCapabilities.Has(servicedef.CapabilityServerSide):
 		fmt.Println("Running server-side SDK test suite")
-		results = sdktests.RunServerSideTestSuite(harness, params.filters.Match, testLogger)
-	case capabilities.Has(servicedef.CapabilityClientSide):
+		allCapabilities = sdktests.AllImportantServerSideCapabilities()
```
eli-darkly, line 69: I added the line about `allCapabilities` to fix a desired behavior that wasn't happening: it is supposed to tell you at the top of the test output whether tests will be skipped because of a missing capability. This requires comparing the list of capabilities supported by the test service to a list of all the ones we would like them to implement.

framework/ldtest/junit_test_logger.go
```diff
@@ -0,0 +1,206 @@
...
+	status.skipped = ldvalue.NewOptionalString(reason)
+	j.tests[id.String()] = status
+}
+
+func (j *JUnitTestLogger) EndLog(results Results) error {
+	fmt.Printf("Writing JUnit data to %s\n", j.filePath)
+
+	var doc jUnitXMLDocument
+
+	properties := []jUnitXMLProperty{
```
eli-darkly, line 126: I think that in the context of actual JUnit, this "properties" list would represent Java system properties that are part of the test run configuration. I've made up some property names here just to capture some of the input parameters that affect the test harness .

keelerm84 (approved):

framework/ldtest/filter.go
```diff
@@ -121,3 +121,26 @@ func PrintFilterDescription(filters RegexFilters, allCapabilities []string, supp
...
+	if filters.MustNotMatch.IsDefined() {
+		parts = append(parts, fmt.Sprintf("must not match %s", filters.MustNotMatch))
+	}
+	if len(parts) == 0 {
+		return "none"
+	}
+	return strings.Join(parts, ", ")
+}
+
+func autoEscapeTestRegex(pattern string) string {
```
eli-darkly, line 139: I added this logic because Matthew very reasonably tried to use a filter expression like `-run 'evaluation/param/general (secondary)'` and got unexpected results due to the regex behavior of parens.

keelerm84 (approved):

PR #12, opened 2022-01-25
implement big segments tests

eli-darkly:
This PR adds a test suite for the non-database-specific components of Big Segments, for SDKs that have that feature.

There are two parts to Big Segment support in a server-side SDK:
* The Big Segment store, which is an integration with a specific database, implementing a standard query API with two methods, "get metadata" and "get user membership" (as defined in the internal spec "Big (Unbounded) Segments (SDK team focused) Spec"). It is normally implemented in the same place as the persistent feature store integration for the same database. The SDK test harness cannot test this, because testing it requires a real database.
* The core logic, which is all the Big Segments-related logic in the SDK that is _not_ specific to a database. This is described in two internal specs: the one mentioned above, and also the parts of the "Feature flag evaluation algorithm" spec that deal with Big Segments.

The new contract tests test the core logic. The test harness creates a simulation of a Big Segment store, implemented as a simple REST service. It provides the base URI of this service as part of the test service configuration. The test service is responsible for creating its own implementation of the Big Segment store API, and for each method in that API, it makes a corresponding call to the simulated big segment store callback service, passing whatever test data (or error) was provided by the test harness back to the SDK. The test harness then verifies that the SDK behaves the way we would expect it to if a real database returned those results, in terms of using the results in evaluations, caching the results, reporting errors, etc.

The test coverage is nearly equivalent to the unit tests that we currently have for this feature in the Go and .NET SDKs. It is testing at a higher level than the unit tests, in that it operates only through the public SDK API rather than testing the underlying lower-level components, but given the fairly simple nature of the interactions between the SDK and the store component, that should be adequate.

One thing that's currently missing is the ability to test the _status listener_ mechanism that the SDKs generally provide— a push notification system for detecting changes in the store status. It should be possible to add tests for this with a similar callback system, but I'll save that for another PR. Once we have that, I believe we will have fully replicated the current test coverage in a cross-platform way, and we can consider simplifying the per-SDK unit tests in this area in the future.

### DIFF
--- a/framework/ldtest/test_scope.go
+++ b/framework/ldtest/test_scope.go
@@ -81,8 +81,8 @@ func (t *T) run(action func(*T)) {
 		if t.failed {
 			t.env.results.Failures = append(t.env.results.Failures, result)
 		}
-		for _, cleanupFn := range t.cleanups {
-			cleanupFn()
+		for i := len(t.cleanups) - 1; i >= 0; i-- {
+			t.cleanups[i]()
 		}
 	}()
 
--- a/mockld/callback_helpers.go
+++ b/mockld/callback_helpers.go
@@ -1,0 +1,60 @@
+package mockld
+
+import (
+	"bytes"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/launchdarkly/sdk-test-harness/framework"
+
+	"github.com/gorilla/mux"
+)
+
+type callbackService struct {
+	router *mux.Router
+	logger framework.Logger
+	name   string
+}
+
+func newCallbackService(logger framework.Logger, name string) *callbackService {
+	router := mux.NewRouter()
+	c := &callbackService{router: router, logger: logger, name: name}
+	router.HandleFunc("/", c.close).Methods("DELETE")
+	return c
+}
+
+func (c *callbackService) addPath(path string, handler func(*json.Decoder) (interface{}, error)) {
+	c.router.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+		var requestDecoder *json.Decoder
+		var body []byte
+		if r.Body != nil {
+			body, _ = ioutil.ReadAll(r.Body)
+			_ = r.Body.Close()
+			requestDecoder = json.NewDecoder(bytes.NewBuffer(body))
+		}
+		c.logger.Printf("[%s] got POST %s %s", c.name, path, string(body))
+
+		responseValue, err := handler(requestDecoder)
+		if err != nil {
+			w.Header().Set("content-type", "text/plain")
+			w.WriteHeader(http.StatusInternalServerError)
+			_, _ = w.Write([]byte(err.Error()))
+			c.logger.Printf("[%s] responded with 500 - %s", c.name, err.Error())
+		} else {
+			w.Header().Set("content-type", "application/json")
+			var respBody []byte
+			w.WriteHeader(http.StatusOK)
+			if responseValue != nil {
+				respBody, _ = json.Marshal(responseValue)
+				_, _ = w.Write(respBody)
+			}
+			c.logger.Printf("[%s] responded with 200 %s", c.name, string(respBody))
+		}
+	})
+}
+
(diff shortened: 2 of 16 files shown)


framework/ldtest/test_scope.go
```diff
@@ -81,8 +81,8 @@ func (t *T) run(action func(*T)) {
 		if t.failed {
 			t.env.results.Failures = append(t.env.results.Failures, result)
 		}
-		for _, cleanupFn := range t.cleanups {
-			cleanupFn()
+		for i := len(t.cleanups) - 1; i >= 0; i-- {
+			t.cleanups[i]()
```
eli-darkly, line 85: This fixes a problem with test cleanup that I noticed in the logs. When a test calls methods like `NewSDKClient`, `NewSDKDataSource`, etc., we use the `t.Defer()` mechanism to ensure that those components are torn down on leaving the test scope. This should behave like Go's `defer` in that the most recently created things are torn down first, but it was doing the opposite, which could cause spurious errors.

mockld/callback_helpers.go
```diff
@@ -0,0 +1,60 @@
...
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/launchdarkly/sdk-test-harness/framework"
+
+	"github.com/gorilla/mux"
+)
+
+type callbackService struct {
```
eli-darkly, line 14: This provides a simple callback service structure that we can reuse for any other instrumented components that the test service might want to provide, if they have multiple endpoints and can take parameters, as in the Big Segment store. For instance, I think we could create cross-platform tests for the persistent data store feature in the same way.

sdktests/server_side_big_segments.go
```diff
@@ -0,0 +1,502 @@
+package sdktests
+
```
eli-darkly, line 2: This file is pretty big— sorry— I tried to factor out boilerplate where I could, but I could probably do more. But I hope the structure is clear at least.

mockld/callback_helpers.go
```diff
@@ -0,0 +1,65 @@
...
+func (c *callbackService) addPath(path string, handler func(*json.Decoder) (interface{}, error)) {
+	c.router.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+		var requestDecoder *json.Decoder
+		var body []byte
+		if r.Body != nil {
+			body, _ = ioutil.ReadAll(r.Body)
+			_ = r.Body.Close()
+			requestDecoder = json.NewDecoder(bytes.NewBuffer(body))
+		}
+		c.logger.Printf("[%s] got POST %s %s", c.name, path, string(body))
```
cwaldren-ld, line 41: It could be useful to surround the body format string with brackets or something to make it clear to people reading the logs the absence of a body.

mockld/callback_helpers.go
```diff
@@ -0,0 +1,65 @@
...
+			c.logger.Printf("[%s] responded with 500 - %s", c.name, err.Error())
+		} else {
+			w.Header().Set("content-type", "application/json")
+			var respBody []byte
+			w.WriteHeader(http.StatusOK)
+			if responseValue != nil {
+				respBody, _ = json.Marshal(responseValue)
+				_, _ = w.Write(respBody)
+			}
+			c.logger.Printf("[%s] responded with 200 %s", c.name, string(respBody))
```
cwaldren-ld, line 57: (ditto)

cwaldren-ld (commented):
Looks great!

cwaldren-ld (approved):

mockld/callback_helpers.go
eli-darkly: Well, the reason I didn't bother doing something like that was, the presence or absence of a request body is just part of the service API spec— that is, a type of request that has parameters will always have them, and a type of request that doesn't have parameters will never have a body— that is not a thing that varies according to some variable SDK condition. In other words, if I'm looking at the log and I see a request was made to `/getMetadata`, I'm not going to be wondering "I don't see a request body, was there one?" because there's no reason for `/getMetadata` to have one.

PR #15, opened 2022-01-27
add event test logic for debug events and contextKind

eli-darkly:
This improves test coverage for SDK event behavior as follows:

* Add tests to verify that `debug` events are generated if and only if they should be generated. This includes testing the somewhat convoluted logic that the SDKs use to check the debug expiration time against both their own current time and the last `Date` header they got from an LD event post, if any (which can't be tested with integration-harness, and is not always possible to test well in the SDK unit tests either).
* For `feature`, `debug`, and `custom` events, we now verify that the SDK puts `"contextKind":"anonymousUser"` in the event for anonymous users.

We expect that all current SDK versions will pass these tests.

This also addresses [sc-139541](https://app.shortcut.com/launchdarkly/story/139541/minor-bugs-in-sdk-test-harness-prerequisite-event-tests).

### DIFF
--- a/mockld/event_data.go
+++ b/mockld/event_data.go
@@ -32,30 +32,21 @@ func (e Event) Kind() string {
 // CanonicalizedJSONString transforms the JSON formatting of the event to make comparisons
 // reliable, as follows: 1. Drop the creationDate property, since its value is unpredictable.
 // 2. Call EventUser.CanonicalizedJSONString on user properties, if any. 3. Ensure that
-// all expected properties, if nullable, do appear in the object when null instead of being
-// omitted.
+// all nullable properties are omitted if their value is null (so we can write more concise
+// test expectations).
 func (e Event) CanonicalizedJSONString() string {
 	b := ldvalue.ObjectBuild()
 	for key, value := range ldvalue.Value(e).AsValueMap().AsMap() {
-		if key == "creationDate" {
+		if value.IsNull() {
+			continue
+		}
+		switch key {
+		case "creationDate":
 			continue // We won't try to make assertions about the timestamp, just omit it
-		}
-		if key == "user" {
+		case "user":
 			b.Set(key, ldvalue.Raw([]byte(EventUser(value).CanonicalizedJSONString())))
-		} else {
+		default:
 			b.Set(key, value)
-		}
-	}
-	var nullableProps []string
-	switch e.Kind() {
-	case "feature", "debug":
-		nullableProps = []string{"userKey", "user", "version", "variation", "reason", "default", "prereqOf"}
-	case "custom":
-		nullableProps = []string{"userKey", "user", "data", "metricValue"}
-	}
-	for _, k := range nullableProps {
-		if !ldvalue.Value(e).GetByKey(k).IsDefined() {
-			b.Set(k, ldvalue.Null())
 		}
 	}
 	return b.Build().JSONString()
@@ -172,3 +163,5 @@ func (u EventUser) CanonicalizedJSONString() string {
 func (u EventUser) AsValue() ldvalue.Value { return ldvalue.Value(u) }
 
 func (u EventUser) GetKey() string { return u.AsValue().GetByKey("key").StringValue() }
+
+func (u EventUser) IsAnonymous() bool { return u.AsValue().GetByKey("anonymous").BoolValue() }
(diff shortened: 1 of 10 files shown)


mockld/event_data.go
```diff
@@ -32,32 +32,23 @@ func (e Event) Kind() string {
 // CanonicalizedJSONString transforms the JSON formatting of the event to make comparisons
 // reliable, as follows: 1. Drop the creationDate property, since its value is unpredictable.
 // 2. Call EventUser.CanonicalizedJSONString on user properties, if any. 3. Ensure that
-// all expected properties, if nullable, do appear in the object when null instead of being
-// omitted.
+// all nullable properties are omitted if their value is null (so we can write more concise
+// test expectations).
```
eli-darkly, line 36: I've reversed the previous behavior of `CanonicalizedJSONString` because I realized it's a lot simpler to just drop nulls, and compare against an expected value where nulls are omitted, than to add missing nulls and compare against an expected value that has nulls.

mockld/events_service.go
```diff
@@ -64,6 +73,14 @@ func (s *EventsService) postEvents(w http.ResponseWriter, r *http.Request) {
 		s.logger.Printf("unable to read request body")
 		return
 	}
+
+	s.lock.Lock()
+	hostTime := s.hostTimeOverride
+	s.lock.Unlock()
+	if !hostTime.IsZero() {
+		w.Header().Set("Date", hostTime.UTC().Format(http.TimeFormat))
```
eli-darkly, line 81: This is the thing we can't simulate with integration-harness. And in some SDKs, we can't simulate it in unit tests either, at least not when we're doing actual HTTP in the tests— some embedded HTTP server frameworks don't allow the `Date` header to be overridden. So in those SDKs, we've only been able to test it by completely mocking out the HTTP client piece, which isn't ideal.

sdktests/custom_matchers.go
```diff
@@ -85,15 +85,11 @@ func EventIsCustomEvent(
...
-		o.Set("userKey", ldvalue.Null())
-	} else {
-		o.Set("user", ldvalue.Null())
-		o.Set("userKey", ldvalue.String(eventUser.GetKey()))
-	}
-	o.Set("data", data)
-	o.Set("metricValue", ldvalue.CopyArbitraryValue(metricValue))
+	setPropertyConditionally(o, inlineUser, "user", eventUser.AsValue())
+	setPropertyConditionally(o, !inlineUser, "userKey", ldvalue.String(eventUser.GetKey()))
+	setPropertyConditionally(o, eventUser.IsAnonymous(), "contextKind", ldvalue.String("anonymousUser"))
```
eli-darkly, line 90: This piece, contextKind, is something we weren't testing before. The rest of the changes in this method are just streamlining how we handle nulls in the expected data.

sdktests/server_side_events_identify.go
```diff
@@ -22,15 +22,17 @@ func doServerSideIdentifyEventTests(t *ldtest.T) {
...
-		user := users.NextUniqueUser()
-		client.SendIdentifyEvent(t, user)
-		client.FlushEvents(t)
-		payload := events.ExpectAnalyticsEvents(t, defaultEventTimeout)
-		m.In(t).Assert(payload, m.Items(
-			EventIsIdentifyEvent(mockld.SimpleEventUser(user)),
-		))
-	})
+	for _, isAnonymousUser := range []bool{false, true} {
+		t.Run(selectString(isAnonymousUser, "anonymous user", "non-anonymous user"), func(t *ldtest.T) {
```
cwaldren-ld, line 26: This is just an idea, but I wonder once we have access to generics if we can make this less verbose. Something like:
```golang
func RunScenarios[Value any](t *ldtest.T, domain map[string]Value, do func (t *ldtest.T, value Value)) {
     for label, value := range domain {
          t.Run(label, func(t *ldtest.T) { do(t, value) })
     }
}

func TestXYZ(t *ldtest.T) {
     RunScenarios(t, 
          map[string]bool{"anonymous": true, "non-anonymous" : false}, 
          func(t *ldtest.T, value bool) {
               //do stuff with value
          }
     )
}
````
I'm sure there will be a bunch of generics-based testing libraries in the coming months 😵‍💫

cwaldren-ld (approved):

sdktests/server_side_events_identify.go
eli-darkly: Yeah, even though it'll be a while before we can use generics in the SDK or in any support libraries that get compiled with the SDK, I'm hoping that we can start leveraging them in stand-alone code like this.

PR #16, opened 2022-01-27
Allow passing -skip parameters to run.sh

keelerm84:
Using the standard Makefile setup, if I try to pass multiple -skip options to the run script, it fails to recognize them all.

### DIFF
--- a/downloader/run.sh
+++ b/downloader/run.sh
@@ -51,4 +51,4 @@ if [ ! -x "${EXECUTABLE}" ]; then
   tar -xf "${TEMP_DIR}/archive.tar.gz" -C "${TEMP_DIR}"
 fi
 
-bash -c "${EXECUTABLE} $PARAMS"
+sh -c "${EXECUTABLE} $PARAMS"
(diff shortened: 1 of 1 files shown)


downloader/run.sh
```diff
@@ -51,4 +51,4 @@ if [ ! -x "${EXECUTABLE}" ]; then
   tar -xf "${TEMP_DIR}/archive.tar.gz" -C "${TEMP_DIR}"
 fi
 
-"${EXECUTABLE}" $PARAMS
+bash -c "${EXECUTABLE} $PARAMS"
```
eli-darkly, line 54: ........I don't get it. Why does this make a difference in terms of allowing multiple -skip parameters? Can you say more about the failure mode you saw?

keelerm84: I am using [this Makefile](https://github.com/launchdarkly/rust-server-sdk-private/blob/mk/sc-139773/add-contract-tests/Makefile). If you run this against the master version of this script, you will see output like

```
Some tests will be skipped based on the filter criteria for this test run:
  skip any matching "'streaming/retry"
```

So it isn't picking up the other `-skip` options. Looks like the spaces in the test names are causing some escaping issues.

eli-darkly: OK, this may be nitpicking, but it sounds to me like the issue is not really that there are multiple `-skip` options per se, it's that parameter values containing spaces are being treated as multiple parameters. Passing multiple parameters definitely does work, no matter what they are, as long as there aren't spaces. I'm still a little confused as to why the single quotes are getting lost, though.

eli-darkly (approved):

PR #17, opened 2022-01-27
add NonCritical option for tests + make some of the "all flags" assertions non-critical

eli-darkly:
See story, plus the new docs text in this PR.

Along with implementing this feature, I made use of it in a test change as a proof of concept. The tests I added in https://github.com/launchdarkly/sdk-test-harness/pull/5 had strict expectations about the SDK stripping out null-valued properties from the JSON data. That's not something we ever really spec'd, even if it's generally desirable— and, because it was lumped in with the overall equality assertions that the tests were making, if an SDK did not strip the null values it would show up as a whole lot of "all flags" tests being broken, rather than showing more clearly that everything else works except for that. So, I changed the existing tests to be less strict in that regard, and added a separate non-critical test for just the null-stripping behavior. I'll do a follow-up PR that adds something similar for null values in events.

### DIFF
--- a/docs/writing_tests.md
+++ b/docs/writing_tests.md
@@ -61,7 +61,7 @@ There are several ways to do this. When using `testify/assert` or `testify/requi
 
 Also, using the `matchers` API sometimes results in more helpful error messages. For instance, if you want to assert that every element in a slice has a certain property, using `matchers.Items` or `matchers.ItemsInAnyOrder` along with a matcher for the desired property is better than doing an equality test on the whole slice against the entire expected data set, because it won't care about any irrelevant properties, and because the property matcher will automatically give you an appropriate failure message. It's also better than doing a `for` loop over the slice and doing an assertion on each element, because you won't have to construct your own message to explain that the failure is for item 0, item 1, etc.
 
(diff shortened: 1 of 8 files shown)


docs/writing_tests.md
```diff
@@ -49,14 +49,30 @@ See documentation comments for a full description of the available API. Here is
...
+
+Also, using the `matchers` API sometimes results in more helpful error messages. For instance, if you want to assert that every element in a slice has a certain property, using `matchers.Items` or `matchers.ItemsInAnyOrder` along with a matcher for the desired property is better than doing an equality test on the whole slice against the entire expected data set, because it won't care about any irrelevant properties, and because the property matcher will automatically give you an appropriate failure message. It's also better than doing a `for` loop over the slice and doing an assertion on each element, because you won't have to construct your own message to explain that the failure is for item 0, item 1, etc.
+
+Values in failure output may be formatted differently in `matchers`versus `testify`: for instance, values of types that have a JSON representation are shown as JSON in `matchers`, rather than using the result of `fmt.Sprintf("%v", value)` as `testify` does in most cases. The easiest way to see which kind of output would be most helpful for a particular test is to cause a deliberate failure in the test and see what you get.
```
keelerm84, line 64:
```suggestion
Values in failure output may be formatted differently in `matchers` versus `testify`: for instance, values of types that have a JSON representation are shown as JSON in `matchers`, rather than using the result of `fmt.Sprintf("%v", value)` as `testify` does in most cases. The easiest way to see which kind of output would be most helpful for a particular test is to cause a deliberate failure in the test and see what you get.
```

keelerm84 (approved):

PR #18, opened 2022-01-27
fix AllFlagsState "detail only for tracked flags" expectation

eli-darkly:
I incorrectly had this test asserting that, if the "detail only for tracked flags" option was used with AllFlagsState, and a flag did not have event tracking or debugging enabled, then the `reason` property should be omitted but the `version` property should still be sent. Actually the intention was for both `reason` and `version` to be omitted— which is how we implemented it in nearly all of the SDKs, but the Go SDK was an outlier and I had based the test on that.

Specifically the idea behind this option is, if an application is passing bootstrap data from the back end to the JS SDK, and they have a ton of client-side flags, they would like the data to be less huge. As long as the front-end code does not care about evaluation reasons (that is, they're never calling `variationDetail`), the reasons will generally be irrelevant, except that if you're also using experimentation or event tracking/debugging, then you _do_ want to see reasons in the detailed `feature` events that it will produce. But for any flags where those things don't apply, `feature` events aren't produced, just summary events. Summary events do include a `version` property but it's not terribly relevant in most cases, so we decided we could do without it.

### DIFF
--- a/sdktests/server_side_eval_all_flags.go
+++ b/sdktests/server_side_eval_all_flags.go
@@ -8,6 +8,7 @@ import (
 	"github.com/launchdarkly/sdk-test-harness/mockld"
 	"github.com/launchdarkly/sdk-test-harness/servicedef"
 
+	m "github.com/launchdarkly/go-test-helpers/v2/matchers"
 	"gopkg.in/launchdarkly/go-sdk-common.v2/ldtime"
 	"gopkg.in/launchdarkly/go-sdk-common.v2/lduser"
 	"gopkg.in/launchdarkly/go-sdk-common.v2/ldvalue"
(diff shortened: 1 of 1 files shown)


sdktests/server_side_eval_all_flags.go
```diff
@@ -292,7 +295,7 @@ func doServerSideAllFlagsDetailsOnlyForTrackedFlagsTest(t *ldtest.T) {
 		},
 		"$valid": true
 	}`
-	assert.JSONEq(t, expectedJSON, string(resultJSON))
+	m.In(t).Assert(resultJSON, m.JSONStrEqual(expectedJSON))
```
eli-darkly, line 298: I switched to using the `matchers` version of this assertion because I like the output better. For instance, if we see a property we didn't expect, `assert.JSONEq` will print something like this:

```
  Not equal:
  expected: map[string]interface {}{"$flagsState":map[string]interface {}{"flag1":map[string]interface {}{"variation":1}, "flag2":map[string]interface {}{"reason":map[string]interface {}{"kind":"OFF"}, "trackEvents":true, "variation":2, "version":200}, "flag3":map[string]interface {}{"debugEventsUntilDate":1.6433153693e+12, "reason":map[string]interface {}{"kind":"OFF"}, "variation":3, "version":300}}, "$valid":true, "flag1":"value1", "flag2":"value2", "flag3":"value3"}
  actual  : map[string]interface {}{"$flagsState":map[string]interface {}{"flag1":map[string]interface {}{"variation":1, "version":100}, "flag2":map[string]interface {}{"reason":map[string]interface {}{"kind":"OFF"}, "trackEvents":true, "variation":2, "version":200}, "flag3":map[string]interface {}{"debugEventsUntilDate":1.6433153693e+12, "reason":map[string]interface {}{"kind":"OFF"}, "variation":3, "version":300}}, "$valid":true, "flag1":"value1", "flag2":"value2", "flag3":"value3"}
  Diff:
  --- Expected
  +++ Actual
  @@ -2,4 +2,5 @@
  (string) (len=11) "$flagsState": (map[string]interface {}) (len=3) {
  -  (string) (len=5) "flag1": (map[string]interface {}) (len=1) {
  -   (string) (len=9) "variation": (float64) 1
  +  (string) (len=5) "flag1": (map[string]interface {}) (len=2) {
  +   (string) (len=9) "variation": (float64) 1,
  +   (string) (len=7) "version": (float64) 100
  },
```

But with `matchers.JSONStrEqual` you get this:
```
  JSON values at "$flagsState"."flag1"."version": expected = <absent>, actual = 100
  full value was: {"$flagsState":{"flag1":{"variation":1,"version":100},"flag2":{"reason":{"kind":"OFF"},"trackEvents":true,"variation":2,"version":200},"flag3":{"debugEventsUntilDate":1643315408682,"reason":{"kind":"OFF"},"variation":3,"version":300}},"$valid":true,"flag1":"value1","flag2":"value2","flag3":"value3"}
```

keelerm84 (approved):

PR #20, opened 2022-01-28
big refactoring & improvement of event tests to make them clearer & more specific

eli-darkly:
Sorry about the size of this PR, but I think it will make the code easier to maintain in the future.

When I wrote these tests before, I was still not sure of the best way to write the assertions, and since the matchers library provides several ways to do similar things, my approach was kind of all over the place in each test. This PR makes the usage a lot more straightforward, and also more specific to what we are actually trying to test:

* Instead of building a JSON object representing the expected event (which required a "canonicalize" step to deal with unimportant differences), and then doing an equality test against it, we now write out the individual property tests we care about.
* In a few places it's still desirable to do an equality test on a JSON object, but I'm now consistently using the matchers library's `JSONEqual` or `JSONStrEqual` methods instead of `assert.JSONEq` for this. They both test deep equality, but the failure output from `assert.JSONEq` is both verbose and (to me) a bit confusing (because they do it by converting the JSON to a Go `map[string]interface{}` structure, printing that in a one-line-per-field format, and then diffing the output) whereas the other ones will just tell you things like "for JSON property 'x'.'y', expected = true, actual = <absent>".
* That allows tests to be much more specific regarding what they care about. If the test expectation is "we expect to see an index event for the same user key generated when a custom event is added", then that test should not have expectations about the specific properties of the `user` object within the index event, other than the user key; all such checks should be in tests designated as being about user properties. The rationale is that if the SDK has a bug causing user properties to be wrong, they will likely be wrong in every event, so if the tests were all doing equality assertions on the whole user object then we would see a confusingly large number of failures even though the other behavior wasn't broken.
* I've made more of an effort to avoid using assertions with non-obvious behaviors. For instance, sometimes I was using a `MapOf` method that took a list of key-value pairs, and besides checking those values, it also verified that no other unexpected keys appeared... but that wasn't obvious if you were looking at the test logic, so instead I added an assertion called `JSONPropertyKeysCanOnlyBe`.

I tested these changes first by checking that the tests that were passing still pass, and then by deliberately breaking particular things in a local copy of an SDK and observing that I got more specific failures


mockld/event_data.go
```diff
@@ -17,41 +13,10 @@ type Event ldvalue.Value
 // Events is an array of events. This specialized type provides helper methods.
 type Events []Event
 
-// EventUser is a JSON representation of a user in an event. LaunchDarkly's JSON schema for users
-// in events is slightly different than users as SDK inputs.
-type EventUser ldvalue.Value
```
eli-darkly, line 22: There's no longer a good reason to have this specialized type, so I've removed a lot of related stuff to simplify things. I was using it mostly for two reasons: 1. in functions like `EventIsFeatureEvent` that took user properties as a parameter, where `lduser.User` wouldn't have been right because that is the input state of the user and not the _event_ representation of a user, and 2. to provide the "canonicalize" logic which was only needed because I was doing equality tests on the whole event.

mockld/event_data.go
```diff
@@ -73,95 +38,3 @@ func (es Events) JSONString() string {
 	data, _ := json.Marshal(es)
 	return string(data)
 }
-
-func SimpleEventUser(user lduser.User) EventUser {
-	// This function can be used when we're not testing anything related to private attributes
-	return ExpectedEventUserFromUser(user, servicedef.SDKConfigEventParams{})
-}
-
-func ExpectedEventUserFromUser(user lduser.User, eventsConfig servicedef.SDKConfigEventParams) EventUser {
```
eli-darkly, line 82: This is replaced by `eventUserMatcher` in `server_side_events_users.go`, which uses composable matchers instead of building a JSON object. I figured it made more sense for the logic to be in that file because now it's _only_ the tests in that file that make assertions about specific user properties in events.

sdktests/server_side_eval.go
```diff
@@ -95,7 +95,7 @@ func RunParameterizedServerSideClientNotReadyEvalTests(t *ldtest.T) {
 			t.Run("evaluate flag without detail", func(t *ldtest.T) {
 				result := client.EvaluateFlag(t, servicedef.EvaluateFlagParams{
 					FlagKey:      flagKey,
-					User:         &user,
+					User:         user,
```
eli-darkly, line 98: I changed the structs in `servicedef` to take just a User rather than a pointer to a User. The only reason it was a pointer was that I was trying to think ahead to reusing some of this stuff for client-side SDKs, where the user is not a parameter. But I think that was premature— I just don't know enough yet about how these tests will be adapted for client-side SDKs— and it made the calling logic inconvenient.

sdktests/server_side_events_custom.go
```diff
@@ -18,66 +17,57 @@ func doServerSideCustomEventTests(t *ldtest.T) {
 
 	t.Run("data and metricValue parameters", doServerSideParameterizedCustomEventTests)
 
-	t.Run("index events are created for new users if not inlined", func(t *ldtest.T) {
```
eli-darkly, line 21: This is moved to `server_side_events_index.go`, which covers the `index` event behavior in detail. Now, in custom event tests where we expect an index event to be generated but it's not the focus of the test, we just use `IsIndexEvent()` to indicate very generally that the event payload will include one.

sdktests/server_side_events_summary.go
```diff
@@ -51,40 +51,35 @@ func doServerSideSummaryEventBasicTest(t *ldtest.T) {
...
-				m.KV(flag1.Key, m.MapOf(
-					m.KV("default", m.JSONEqual(default1)),
-					m.KV("counters", m.ItemsInAnyOrder(
-						flagCounter("value1a", 0, flag1.Version, 2),
-						flagCounter("value1b", 1, flag1.Version, 1),
-					)),
+		IsIndexEvent(),
+		IsIndexEvent(),
+		IsValidSummaryEventWithFlags(
+			m.KV(flag1.Key, m.MapOf(
```
eli-darkly, line 68: Here I've continued to use the `MapOf` method, which I stopped using in most other cases, because the summary event data structure is one big nested thing, and all of the properties within `features` are predictable and relevant to this test.

sdktests/custom_matchers_events.go
```diff
@@ -0,0 +1,93 @@
...
+
+func HasContextKind(user lduser.User) m.Matcher {
+	if user.GetAnonymous() {
+		return m.JSONProperty("contextKind").Should(m.Equal("anonymousUser"))
+	}
+	return JSONPropertyNullOrAbsent("contextKind")
+}
+
+func HasAnyCreationDate() m.Matcher {
+	return m.JSONProperty("creationDate").Should(ValueIsPositiveNonZeroInteger())
```
keelerm84, line 34: If a value is positive, it is already defined as non zero so you could theoretically simply this name some to ValueIsPositiveInteger. If you wanted a positive or zero value, then we could have ValueIsNonNegativeInteger. But maybe that's too pedantic?

keelerm84 (approved):
I like these changes. I think this is going to make determining failure reasons much easier!

sdktests/custom_matchers_events.go
eli-darkly: You're strictly correct, but non-mathematicians so often conflate "positive" with "nonnegative" that I prefer to spell it out like this for clarity.

PR #28, opened 2022-02-04
Add ability to generate / load suppressions file

cwaldren-ld:
This PR introduces a suppression file which can be generated by the Test Harness at runtime. 

During subsequent runs, the suppression file can be passed into the harness in order to simulate `-skip` behavior for each line in the file.

This allows for gradual compliance with the Test Harness.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	_ "embed" // this is required in order for go:embed to work
 	"fmt"
 	"log"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -29,6 +30,26 @@ func main() {
 		os.Exit(1)
 	}
 
+	if params.suppressions != "" {
+		file, err := os.Open(params.suppressions)
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "Cannot open provided suppression file: ", err)
+			os.Exit(1)
```
cwaldren-ld, line 37: We _might_ be outgrowing this early-out pattern - perhaps we should have something like:
```golang
func main() {
    if err := app.run(); err != nil {
        // print error
        os.Exit(1)
     }
}
```

eli-darkly: Sure, I'd be fine with that.

params.go
```diff
@@ -30,6 +32,8 @@ func (c *commandParams) Read(args []string) bool {
 	fs.BoolVar(&c.debug, "debug", false, "enable debug logging for failed tests")
 	fs.BoolVar(&c.debugAll, "debug-all", false, "enable debug logging for all tests")
 	fs.StringVar(&c.jUnitFile, "junit", "", "write JUnit XML output to the specified path")
+	fs.StringVar(&c.genSuppressions, "gen-suppressions", "", "output suppression file")
+	fs.StringVar(&c.suppressions, "suppressions", "", "input suppression file")
```
eli-darkly, line 36: These parameter names and descriptions are a bit vague; if I just saw this list and hadn't read the code, I would have no idea what they did (especially since the PR doesn't include any additions to the command-line instructions in `docs/`).

How about something more specific to what they are doing, such as:

`-list-failed-tests-to` ... `writes names of failed tests to the specified file, one per line`
`-skip-all-from` ... `skips tests named in the specified file, one per line`

PR #30, opened 2022-02-04
Allow gradual compliance with Test Harness via test failure suppression file

cwaldren-ld:
This PR introduces a simple method to mass-skip tests, allowing an SDK to incrementally comply with the contract test suite.

If the `-record-failures [file-path]` argument is supplied, the Test Harness will record the test IDs of all failures encountered to the specified file.

If `-suppress-failures [file-path]` is supplied, the Test Harness will skip any test that literally matches the string found on each line of the file. 

A common use-case might look like:
```shell
./sdk-test-harness [other args] -record-failures failures.txt
# Now failures.txt contains the name of each failing test.

./sdk-test-harness [other args] -suppress-failures failures.txt
# Now all tests recorded in failures.txt will be skipped.
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -49,7 +49,12 @@ func run(params commandParams) (*ldtest.Results, error) {
 		if err != nil {
 			return nil, fmt.Errorf("cannot open provided suppression file: %v", err)
 		}
-		defer file.Close()
+		closeSuppressionFile := func() {
+			if err := file.Close(); err != nil {
+				fmt.Fprintf(os.Stderr, "Failed to close suppression file: %v\n", err)
+			}
+		}
+		defer closeSuppressionFile()
 		scanner := bufio.NewScanner(file)
 		for scanner.Scan() {
 			escaped := regexp.QuoteMeta(scanner.Text())
@@ -60,7 +65,8 @@ func run(params commandParams) (*ldtest.Results, error) {
 		if err := scanner.Err(); err != nil {
 			return nil, fmt.Errorf("while processing suppression file: %v", err)
 		}
-		_ = file.Close()
+		// Explicit close because this file may be re-opened later on to record an updated suppression file.
+		closeSuppressionFile()
 	}
 
 	mainDebugLogger := framework.NullLogger()
--- a/main.go
+++ b/main.go
@@ -43,8 +43,8 @@ func main() {
 }
 
 func run(params commandParams) (*ldtest.Results, error) {
-	if params.suppressions != "" {
-		file, err := os.Open(params.suppressions)
+	if params.suppressFailures != "" {
+		file, err := os.Open(params.suppressFailures)
 		if err != nil {
 			return nil, fmt.Errorf("cannot open provided suppression file: %v", err)
 		}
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -29,6 +32,36 @@ func main() {
 		os.Exit(1)
 	}
 
+	results, err := run(params)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Test Service Error: %v\n", err)
```
eli-darkly, line 37:
```suggestion
		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
```
"Test service" would refer to the service implementation provided by the SDK, so there are lots of error conditions below that are not a "test service error".

main.go
```diff
@@ -29,6 +32,36 @@ func main() {
...
+		os.Exit(1)
+	}
+}
+
+func run(params commandParams) (*ldtest.Results, error) {
+	if params.suppressFailures != "" {
+		file, err := os.Open(params.suppressFailures)
+		if err != nil {
+			return nil, fmt.Errorf("cannot open provided suppression file: %v", err)
+		}
```
eli-darkly, line 51: Not super important because the program will be terminating anyway, but I think you want `defer file.Close()` here rather than putting `Close` at the end, since you would want it to be closed if there was an early return due to a scanner error.

sdktests/server_side_stream_retry.go
```diff
@@ -122,7 +122,7 @@ func doServerSideStreamRetryTests(t *ldtest.T) {
 		expectNoMoreRequests(t, streamEndpoint)
 	}
 
-	t.Run("retry after I/O error on initial connect", func(t *ldtest.T) {
+	t.Run("retry after IO error on initial connect", func(t *ldtest.T) {
```
eli-darkly, line 125: I don't understand why you felt the need to copyedit this. I/O is a perfectly normal way to write that term in English. The fact that it's written `IO` in various function/type names is irrelevant.

eli-darkly (changes requested):

sdktests/server_side_stream_retry.go
cwaldren-ld: Sorry, shouldn't have added this with no explanatory comment. This isn't arbitrary copy-editing - the framework is unable to suppress these tests since it parses the `/` as part of the test path separator. 

The framework uses the `(t TestID) Plus` method to build up the test ID. It will append `streaming/retry behavior` to `retry after I/O error on reconnect`, the name of the test, resulting in segments:
```
streaming
retry behavior
retry after I/O error on reconnect
```

But when parsing `-skip` arguments, the framework uses ` ParseTestIDPattern` which uses:
```
parts := strings.Split(s, "/")
```
resulting in:
```
streaming
retry behavior
retry after I
O error on reconnect
```
which will not match.

An alternative to all of this would be treating the suppression test IDs as literal strings, and store them in a different variable altogether instead of using the `skip` API. Then we'd need to figure out how `skip/run` interacts with the suppressions, which is why I originally used the `skip` method.

eli-darkly: Oh! Ha! Yeah, sorry, a slash in a test name is problematic. In fact it might even be worth enforcing this in the framework - like, have it actually fail the test as soon as `t.Run` is called with a name like that, since it's extremely likely to cause problems.

eli-darkly: But I can fix that separately. This change is fine. The only one of my comments here that I care about is the misleading "test service error" message.

eli-darkly (approved):

PR #31, opened 2022-02-07
Add square brackets to regex autoescape list

cwaldren-ld:
**Requirements**

- [x] I have added test coverage for new or changed functionality
- [x] I have followed the repository's [pull request submission guidelines](../blob/master/CONTRIBUTING.md#submitting-pull-requests)
- [ ] I have validated my changes against all supported platform versions

**Related issues**

Relates to https://github.com/launchdarkly/sdk-test-harness/pull/30 since many tests containing square brackets may need to be skipped.

**Describe the solution you've provided**

Currently the harness is unable to skip tests that contain square brackets in the ID (e.g. `events/user properties/inlineUsers=true, globally-private=[firstName], user-private=[lastName]/identify event`), because brackets constitute a regex character class.

With this change, the harness is now able to match against these tests.

**Describe alternatives you've considered**

Follows existing pattern established with parenthesis.

### DIFF
--- a/framework/ldtest/filter_test.go
+++ b/framework/ldtest/filter_test.go
@@ -94,4 +94,6 @@ func TestRegexFilters(t *testing.T) {
 func TestAutoEscapeTestRegex(t *testing.T) {
 	assert.Equal(t, "hello \\(yes\\)", autoEscapeTestRegex("hello (yes)"))
 	assert.Equal(t, "hello \\(yes\\)", autoEscapeTestRegex("hello \\(yes\\)"))
+	assert.Equal(t, "hello \\[yes\\], a \\[test\\]", autoEscapeTestRegex("hello [yes], a [test]"))
+	assert.Equal(t, "hello \\[yes\\], a \\[test\\]", autoEscapeTestRegex("hello \\[yes\\], a \\[test\\]"))
 }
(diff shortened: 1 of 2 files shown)


framework/ldtest/filter_test.go
```diff
@@ -94,4 +94,6 @@ func TestRegexFilters(t *testing.T) {
 func TestAutoEscapeTestRegex(t *testing.T) {
 	assert.Equal(t, "hello \\(yes\\)", autoEscapeTestRegex("hello (yes)"))
 	assert.Equal(t, "hello \\(yes\\)", autoEscapeTestRegex("hello \\(yes\\)"))
+	assert.Equal(t, "hello \\[yes\\], a \\[test\\]", autoEscapeTestRegex("hello [yes], a [test]"))
+	assert.Equal(t, "hello \\[yes\\], a \\[test\\]", autoEscapeTestRegex("hello \\[yes\\], a \\[test\\]"))
```
cwaldren-ld, line 98: Wanted to ensure the escape would happen to all of the matched characters in a string, hence the repeat. Could add this to the `()` tests as well.

eli-darkly:
I think I would prefer to just change the way the test is named in this case. Square brackets are a fairly useful feature in regex search; parens, much less so.

PR #33, opened 2022-02-08
don't worry about ordering of events as long as all expected events are present

eli-darkly:
The order in which events appear in a payload isn't significant to LaunchDarkly (if we wanted to look at them in chronological order we would go by the timestamp property) and isn't part of the analytics events spec, so these tests should use the `ItemsInAnyOrder` matcher instead of `Items` whenever they're looking at a list of more than one item.


keelerm84 (approved):
I appreciate the fast fix!

kinyoklion (approved):
Makes erlang happy.

PR #34, opened 2022-02-08
Refactor loading of suppression file into method

cwaldren-ld:
I jumped the gun slightly on merging https://github.com/launchdarkly/sdk-test-harness/pull/30. 

When I modified the PR to defer the suppression file close operation, I didn't realize that it would output an error (because there's also an explicit close.)

In this PR I've implemented the defer correctly by refactoring into a method, so all exit points will execute `Close` once and only once. This is cleaner overall.

The linter gave me an error:
```
main.go:135:2: G307: Deferring unsafe method "Close" on type "*os.File" (gosec)
```
Which led me down a rabbit hole: https://giters.com/securego/gosec/issues/453
To: https://github.com/securego/gosec/issues/579

False positive was fixed upstream here: https://github.com/securego/gosec/releases/tag/v2.9.3
Golang-ci fixed here: https://github.com/golangci/golangci-lint/releases/tag/v1.44.0

So I've upgraded `golang-ci` to v1.44.0.

### DIFF
--- a/main.go
+++ b/main.go
@@ -132,6 +132,8 @@ func loadSuppressions(params *commandParams) error {
 	if err != nil {
 		return fmt.Errorf("cannot open provided suppression file: %v", err)
 	}
+	// #nosec G307 -- This is a read-only file so Close failing will not result in data loss.
+	// Additionally, this is a false positive since we are checking the error code.
 	defer func() {
 		if err := file.Close(); err != nil {
 			fmt.Fprintf(os.Stderr, "Failed to close suppressions file: %v\n", err)
--- a/main.go
+++ b/main.go
@@ -45,7 +45,7 @@ func main() {
 
 func run(params commandParams) (*ldtest.Results, error) {
 	if params.suppressFailures != "" {
-		if err := loadSuppressions(params); err != nil {
+		if err := loadSuppressions(&params); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -145,3 +126,26 @@ func run(params commandParams) (*ldtest.Results, error) {
 
 	return &results, nil
 }
+
+func loadSuppressions(params *commandParams) error {
```
eli-darkly, line 130: What's the purpose of passing `params` as a pointer instead of a value? You're not modifying anything in it.

main.go
```diff
@@ -145,3 +126,26 @@ func run(params commandParams) (*ldtest.Results, error) {
...
+	file, err := os.Open(params.suppressFailures)
+	if err != nil {
+		return fmt.Errorf("cannot open provided suppression file: %v", err)
+	}
+	defer func() {
+		if err := file.Close(); err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to close suppressions file: %v\n", err)
```
eli-darkly, line 137: Despite what the articles say about the reasons to always check the return value of Close(), I honestly feel this is overkill. I mean... if there does happen to be a freak occurrence of the last write failing, there is absolutely nothing we can do about it and the error message isn't going to help us at all, we still have an incomplete file. But printing this doesn't hurt anything, I guess.

eli-darkly (approved):

main.go
cwaldren-ld: Via `(t *TestIDPatternList).Set()` (line 143) it's modifying the `filters.MustNotMatch` member of `params`. That  does an append to the slice.

eli-darkly: 🤦  Sorry, apparently I can't read

PR #35, opened 2022-02-09
Rename -suppress-failures to -skip-from

cwaldren-ld:
Eli brought up the good point that the use-case for `-suppress-failures` is both more general, and different, than the name implies.

More general:
- You might be skipping tests not due to failure, but for other reasons (for example flakiness)

Different:
- It's not that the tests referenced by `-suppress-failures` are run and then the results suppressed; they are skipped entirely.

To head-off confusion, I've renamed the argument to `-skip-from` & updated associated documentation.


eli-darkly (approved):
Thanks for humoring my nitpicks.

PR #37, opened 2022-03-01
add contract tests for tags

eli-darkly:
This adds some new tests which will only be done if the test service includes `"tags"` in its capability list.

The tests cover the formatting of the X-LaunchDarkly-Tags header and the validation rules. Although in principle the SDK should always be using the same set of computed default headers (Authorization, User-Agent, X-LaunchDarkly-Wrapper, etc.) in HTTP requests from every component, the tests check this by looking at both stream requests and event posts— this would also apply to polling requests, but the test harness doesn't currently have the ability to test polling mode.

The tests do not verify that the SDK logs a warning message for bad tags. Currently the test harness doesn't have the ability to query log output; we may want to add that at some point, but in any case we haven't standardized what the log messages would be.

### DIFF
--- a/docs/service_spec.md
+++ b/docs/service_spec.md
@@ -80,7 +80,9 @@ A `POST` request indicates that the test harness wants to start an instance of t
   * `bigSegments` (object, optional): Enables and configures Big Segments. Properties are:
     * `callbackUri` (string, required): The base URI for the big segments store callback fixture. See [Callback fixtures](#callback-fixtures).
     * `userCacheSize`, `userCacheTimeMs`, `statusPollIntervalMS`, `staleAfterMs`: These correspond to the standard optional configuration parameters for every SDK that supports Big Segments.
-  * `tags` (object, optional): If specified, this is an object where each key is a tag key, and each value is an array of strings that are the tag values for that key (always an array even if there is only one value).
+  * `tags` (object, optional): If specified, this has options for metadata/tags (that is, values that are translated into an `X-LaunchDarkly-Tags` header):
+    * `applicationId` (string, optional): If present and non-null, the SDK should set the "application ID" property to this string.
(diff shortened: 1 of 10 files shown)


keelerm84 (approved):

eli-darkly:
I've pushed some changes to match the latest spec:

* The expected delimiter between name-value pairs is now a space instead of a colon.
* Instead of an arbitrary `map[string][]string`, we now pass specific individual parameters to the test service, because we expect the SDKs to have their own property names or setter methods for "application ID" and "application version".

docs/service_spec.md
```diff
@@ -74,6 +80,9 @@ A `POST` request indicates that the test harness wants to start an instance of t
   * `bigSegments` (object, optional): Enables and configures Big Segments. Properties are:
     * `callbackUri` (string, required): The base URI for the big segments store callback fixture. See [Callback fixtures](#callback-fixtures).
     * `userCacheSize`, `userCacheTimeMs`, `statusPollIntervalMS`, `staleAfterMs`: These correspond to the standard optional configuration parameters for every SDK that supports Big Segments.
+  * `tags` (object, optional): If specified, this has options for metadata/tags (that is, values that are translated into an `X-LaunchDarkly-Tags` header):
+    * `applicationId` (string, optional): If present and non-null, the SDK should set the "application ID" property to this string.
```
kinyoklion, line 84: This is `applicationID` (capital D) in the implementatiion.

kinyoklion (changes requested):

docs/service_spec.md
eli-darkly: Whoops - it's the implementation that is wrong. The JSON schema here is using JS conventions, not Go conventions.

kinyoklion (approved):

PR #38, opened 2022-03-03
Add custom attributes when testing events

keelerm84:
Custom attribute handling and filtering is something an SDK is likely to
get wrong. Adding custom attributes to the user property tests ensures
we have basic coverage of this functionality.


keelerm84:
Verified the go sdk works against these changes. Verified the rust one failed and then passed post fix.

eli-darkly (approved):

PR #39, opened 2022-03-04
bump major version to 2

eli-darkly:
The current major feature work will result in SDKs with different enough behavior (features are being both added and removed) that it's best to go to a new major version of sdk-test-harness, rather than try to make all of that behavior conditional on the SDKs' reported capabilities.

The other, more incremental feature work that's currently ongoing (tags) will go into the existing v1.


keelerm84:
Seems like a reasonable thing to do to me. I'm curious -- can go not handle versioning without specifying an alternative branch name? I know Go has some weird versioning history but I'm not well versed enough to understand why that sort of change is required, but definitely curious!

cwaldren-ld (approved):

PR #40, opened 2022-03-04
remove support for inlined users and alias events

eli-darkly:
The only part this that's not straightforward removals is that in the custom event tests, we were previously using `inlineUsersInEvents` just as a convenience so we would not get any index events in the output... so, I just had to change the test assertions slightly to expect an index event as well as the custom event.


sdktests/server_side_events_custom.go
```diff
@@ -20,64 +20,43 @@ func doServerSideCustomEventTests(t *ldtest.T) {
...
-									IsCustomEvent(),
-									HasUserKeyProperty(user.GetKey()),
-									HasContextKind(user),
-								),
-							))
-						}
-					})
-				}
+		for _, anonymousUser := range []bool{false, true} {
+			t.Run(selectString(anonymousUser, "anonymous user", "non-anonymous user"), func(t *ldtest.T) {
```
eli-darkly, line 24: This "anonymous vs. non-anonymous" logic will probably go away too in the new context model— its purpose in these tests was to verify that the event's `contextKind` is set properly, which will no longer be a thing.

cwaldren-ld (approved):

keelerm84 (approved):

PR #43, opened 2022-03-04
(#1) add tests for building & converting context type in strongly-typed SDKs

eli-darkly:
This adds contract test coverage for the behavior of context builders and context JSON converters in strongly-typed SDKs. In terms of JSON representations, this is for the regular schema of context properties— not a context in event data (there will be separate tests for that).

There are two general kinds of new tests here. First, assuming every strongly-typed(*) SDK that supports the new model has some kind of context builder, the test harness will pass some parameters to the test service, like key, name, etc., and say "Build a context out of this, convert it to JSON and send me the JSON. Or, if there's a validation error, tell me that." So this verifies both the builder behavior and the JSON marshaling behavior.

Second, the test harness will pass a JSON representation to the test service and say "Parse this into a context, and then re-serialize that context to JSON and send it back. Or, if parsing or validation fails, tell me that." This verifies both the unmarshaling logic(**) and the logic for converting an old-style user into a context.

(* In the past, the "strongly-typed" category just referred to whether _flag variations_ are strongly typed. Here I'm assuming that that also means the SDK has a specific Context type with a builder and a parser, as opposed to a context just being a JSON object in the SDK. If that ever changes— for instance, if we started using a specific JS class called Context in the Node SDK, even though flag variations are still loosely typed— then we can make these tests conditional on some other capability value.)

(** Unmarshaling users/contexts from JSON isn't a likely use case for customers, but _we_ use it, and just for the general principle of interoperability it's good to verify that all the SDKs understand the schema in the same way, so applications on different platforms could pass user properties to each other in JSON if they wanted to.)

These tests also enforce that the representations are minimal: that is, for instance, we should always omit an empty private attributes list or an empty `_meta` object or a false `transient` value. We will have this requirement for contexts in event data too.

https://app.shortcut.com/launchdarkly/story/144712/sdk-contract-tests-for-context-user-serialization-deserialization


sdktests/sdk_context_type.go
```diff
@@ -0,0 +1,291 @@
...
+			}
+		})
+	})
+
+	t.Run("invalid old user", func(t *ldtest.T) {
+		inputs := []string{
+			`{}`,
+			`{"key": true}`,
+			`{"key": "a", "anonymous": 3}`,
+			`{"key": "a", "anonymous": 3}`,
```
louis-launchdarkly, line 277: Do you mean `{"key": "a", "anonymous": {}}` in this case?

louis-launchdarkly (approved):
Other than the one comment that I have, the Test harness functionality for testing context is comprehensive from what I can tell.

sdktests/sdk_context_type.go
eli-darkly: No, it's just an accidentally duplicated line. Thanks.

eli-darkly: Oh, actually it was supposed to be `secondary`— that was the other built-in one that wasn't covered here.

PR #45, opened 2022-03-08
(#3) migrate to new (prerelease) major versions of SDK packages, use ldcontext

eli-darkly:
This replaces `go-sdk-common.v2` with `go-sdk-common.v3` (using the latest prerelease tag) and changes all usage of `lduser.User` to `ldcontext.Context`.

Nearly all of this is just straight renames; there is no new test functionality related to context features yet (so, for instance, I haven't added any test cases related to multi-kind contexts). I did have to make a few logic changes as noted in PR comments to make the existing tests pass.


sdktests/custom_matchers_events.go
```diff
@@ -23,10 +23,18 @@ func EventHasKind(kind string) m.Matcher {
 	return m.JSONProperty("kind").Should(m.Equal(kind))
 }
 
-func HasContextKeys(user lduser.User) m.Matcher {
-	return m.JSONProperty("contextKeys").Should(m.MapOf(
-		m.KV("user", m.Equal(user.GetKey())),
-	))
+func HasContextKeys(context ldcontext.Context) m.Matcher {
+	var kvs []m.KeyValueMatcher
+	if context.Multiple() {
```
eli-darkly, line 28: There aren't yet any tests that actually use a multi-kind context, but when we do, this logic will apply.

sdktests/sdk_context_type.go
```diff
@@ -185,6 +185,7 @@ func doSDKContextConvertTests(t *ldtest.T) {
 		}
 
 		params := []contextConversionParams{
+			{`{"key": ""}`, `{"kind": "user", "key": ""}`}, // empty key *is* allowed for old user format only
```
eli-darkly, line 188: New test case for verifying the user conversion logic.

sdktests/server_side_events_users.go
```diff
@@ -153,58 +153,57 @@ func doServerSideEventUserTests(t *ldtest.T) {
...
-
-	expected := []m.KeyValueMatcher{
-		m.KV("kind", m.Equal("user")),
+// eventUserMatcher returns a Matcher to verify that a JSON object has the expected properties based on
+// the input context and the events configuration. This mostly means verifying that private attributes
+// behave correctly.
+//
+// Because the rules for private attributes are fairly complicated, we have not reimplemented that logic
+// here. Instead, this function depends on the Go SDK's implementation of event context formatting,
+// treating it as a reference implementation that has been validated by its own thorough tests.
```
eli-darkly, line 162: This is an oversimplified implementation that only works for private attributes that are specified by a simple attribute name; it doesn't work for attribute reference paths. I think probably I'll end up getting rid of this function and just directly declaring what the expected context JSON is as part of the test parameters, but for now, this is the simplest change to let the current test logic work.

cwaldren-ld (approved):

PR #48, opened 2022-03-17
refactor data generators, rename testdata package

eli-darkly:
The previous https://github.com/launchdarkly/sdk-test-harness/pull/44 and https://github.com/launchdarkly/sdk-test-harness/pull/45 had made just the bare minimum changes necessary to let the existing event-related tests pass, but didn't cover all of the new context-related event behavior.

When I started to do more work on the latter, I realized that the UserFactory/ValueFactory generator types I had written for those tests were kind of confusing and inconvenient— especially when it came to tests that had to check different categories of contexts (in the old model, we had to test both anonymous and non-anonymous users). I came up with what I think is a better model. Also, I felt like these types were cluttering up the `sdktests` package a bit and were more closely related to the data-file-loading stuff in the `testdata` package, so I moved them into `testdata`.

However, I also realized that `testdata` was not a good name for a package, because it has special meaning in Go: it'll assume any package with that name is _only_ data and will skip compiling it by default (I hadn't noticed that the unit tests for that package weren't actually running— in fact there were some syntax errors in them!). So I renamed it to `data`.


sdktests/server_side_events_custom.go
```diff
@@ -18,11 +19,10 @@ func doServerSideCustomEventTests(t *ldtest.T) {
...
 
 	t.Run("basic properties", func(t *ldtest.T) {
-		users := NewUserFactory("doServerSideCustomEventTests-")
 		metricValue := 1.0
-		for _, anonymousUser := range []bool{false, true} {
-			t.Run(selectString(anonymousUser, "anonymous user", "non-anonymous user"), func(t *ldtest.T) {
-				user := users.NextUniqueUserMaybeAnonymous(anonymousUser)
+		for _, contexts := range data.NewContextFactoriesForAnonymousAndNonAnonymous("doServerSideCustomEventTests") {
+			t.Run(contexts.Description(), func(t *ldtest.T) {
+				user := contexts.NextUniqueContext()
```
eli-darkly, line 25: Here's the deal with the anonymous/non-anonymous stuff. In the old event model, many events had an optional `contextKind` property which was absent for normal users, and set to `"anonymousUser"` for anonymous users— regardless of what other properties the user might have. So, whenever we were testing one of those events, we wanted to make sure its properties are right for each of those two scenarios.

However, it was a bit clunky to keep writing this very specific logic over and over... and, in the new U2C model, those aren't the test cases we want anyway(*), and there won't necessarily be just two of them. So I felt it was better to generalize the logic so that we just call a generator that returns some ContextFactory instances, each of which has a descriptive name and knows how to create the right kind of contexts, and the test just iterates through those.

(* With U2C, there's no `contextKind` property, instead there is `contextKeys` and the variant behaviors we're interested in have to do with single-kind versus multi-kind— plus, there is some other event behavior (key deduplication) that is slightly different for the "user" kind and other single kinds, so there are really three variants of the behavior.)

sdktests/server_side_events_users.go
```diff
@@ -72,17 +71,18 @@ func doServerSideEventUserTests(t *ldtest.T) {
...
 	defaultValue := ldvalue.String("default")
-	users := NewUserFactory("doServerSideEventUserTests",
-		func(ub lduser.UserBuilder) {
-			ub.FirstName("first").LastName("last").Country("us").
-				Custom("preferredLanguage", ldvalue.String("go")).Custom("primaryLanguage", ldvalue.String("go"))
+	contexts := data.NewContextFactory(
+		"doServerSideEventUserTests",
+		func(b *ldcontext.Builder) {
+			b.SetString("firstName", "first").SetString("lastName", "last").SetString("country", "us").
+				SetString("preferredLanguage", "go").SetString("primaryLanguage", "go")
```
eli-darkly, line 78: This data is a bit obsolete because those attribute names have no special meaning any more, but I'll address that in a bigger update to these tests.

cwaldren-ld (approved):

PR #55, opened 2022-03-25
tolerate variants of redacted attribute name in events

eli-darkly:
If the SDK (or the individual context) is configured to redact the attribute "attr", that could be specified either as "attr" or "/attr", and the SDK is allowed to use either variant when it builds the `redactedAttributes` list. The tests had an overly specific expectation about this, based on how the Go SDK happened to do it.

### DIFF
--- a/sdktests/custom_matchers_events.go
+++ b/sdktests/custom_matchers_events.go
@@ -1,6 +1,8 @@
 package sdktests
 
 import (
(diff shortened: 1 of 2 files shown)


kinyoklion (approved):

kinyoklion:
Looks good, and also works against the node implementation.

PR #56, opened 2022-03-28
add detailed rollout/experiment bucketing tests (v1)

eli-darkly:
This adds more thorough coverage of the behavior of rollouts and experiments in flag evaluations. This PR is specifically for the 1.x version of sdk-test-harness that is used for current GA SDK versions. The equivalent in 2.x, for not-yet-released SDK versions, will be slightly different in terms of context-specific behaviors and other spec changes.

The part that is included in the parameterized evaluation tests, from a YAML file, is pretty basic. The main coverage is implemented in Go code, and included permutations of various factors like:

* is it a rollout or an experiment
* if it's a rollout, is there a secondary key and/or are we bucketing by an attribute other than the key
* if it's an experiment, is there a seed

Unlike the corresponding Go SDK unit tests, the expectations here don't use hard-coded precomputed hash values. Instead, there's an implementation in sdk-test-harness of the hash algorithm, as per the evaluation spec— which is pretty simple once you remove the "which attribute do I bucket by" logic. There are also tests of invalid conditions like "the bucketBy attribute has a value of the wrong type."

Since we don't have direct access to the low-level logic that says what the bucket value is, we test it indirectly like so: if we believe the bucket value should be X, then we set up a rollout/experiment where the first bucket has a weight slightly smaller than X; the second bucket, which is associated with the variation we expect to see, has a very small weight; and the third bucket has a weight that takes up the rest of the full range. While this is not as nice as a direct test of the value (since if there's a problem we won't be able to see directly what the wrong value was, and will have to do some more direct debugging of the SDK), I believe these tests provide more thorough coverage than the current unit tests in any of the SDKs (except maybe for Go, where I believe they are equivalent).


sdktests/server_side_eval_bucketing.go
```diff
@@ -0,0 +1,344 @@
...
+func RunServerSideEvalBucketingTests(t *ldtest.T) {
+	unwantedVar, expectedFallthroughVar, expectedRuleVar, expectedSegmentVar := 0, 1, 2, 3
+	expectedFallthroughValue := ldvalue.String("expected-value-for-fallthrough")
+	expectedRuleValue := ldvalue.String("expected-value-for-rule")
+	expectedSegmentValue := ldvalue.String("expected-value-for-segment-match")
+	unwantedValue, defaultValue := ldvalue.String("wrong"), ldvalue.String("default")
+	matchRuleAttr := "matchrule"
+
+	makeRolloutVariationsToMatch := func(expectedBucketValue float32, desiredVariation int) []ldmodel.WeightedVariation {
+		targetBucketSize := 10 // arbitrary small number (0.01% of 100000) to allow for a tiny bit of rounding error
```
eli-darkly, line 150: This is pretty arbitrary and I'm not sure how much variance we want to tolerate. I mean, theoretically _none_, but since all of the SDKs currently converting integer weights to fractions by dividing by a power of 10, which can't ever be exact in binary floating point, I'm not sure how feasible that would be.

sdktests/helpers.go
```diff
@@ -42,6 +46,46 @@ func basicEvaluateFlag(
...
+
+	hashOutputBytes := sha1.Sum([]byte(hashInput)) //nolint:gosec // this isn't for authentication
+	hexEncodedChars := make([]byte, 64)
+	hex.Encode(hexEncodedChars, hashOutputBytes[:])
+	hash := hexEncodedChars[:15]
+
+	hashVal, _ := strconv.ParseInt(string(hash), 16, 64)
+	var product, result big.Int
+	product.Mul(big.NewInt(hashVal), big.NewInt(100000))
+	result.Div(&product, big.NewInt(0xFFFFFFFFFFFFFFF))
```
eli-darkly, line 85: We've had an ongoing discussion for quite a while about whether it would be best for the SDKs to use integer math instead of floating point in the hash calculation. This is the first time we actually tried doing that— and I realized that of course it can't be done with plain uint64s, because `(hashVal * 100000) / 0xFFFFFFFFFFFFFFF` will overflow in the multiplication step. The quickest solution here was to use Go's `big` types. We can discuss this further within the team if necessary in terms of whether it'd be worthwhile to do anything similar in the actual SDK code.

kinyoklion (commented):
Aside from the ongoing discussion about the method of calculating buckets, I think this looks good. I also don't see a reason we cannot update that method (here) once we reach consensus if needed.

cwaldren-ld (approved):

eli-darkly:
I pushed a change to designate the "secondary key == `""`" subtests as non-critical, because some existing SDKs are known to be out of compliance in this one edge case. When I implement the equivalent of this PR in the v2 sdk-test-harness, I'm planning to remove the `NonCritical` so all U2C-aware SDKs must be consistent on this.

PR #57, opened 2022-03-28
add tests for context kind in segments

eli-darkly:
See PR comments. The behavior for regular segments is pretty straightforward, so those cases are just added to the YAML files for the parameterized tests. For big segments, there's other behavior that we need to cover in Go code.

### DIFF
--- a/sdktests/server_side_big_segments.go
+++ b/sdktests/server_side_big_segments.go
@@ -27,11 +27,11 @@ var bigSegmentsExpectedHash = "CEXjZY7cHJG/ydFy7q4+YEFwVrG3/pkJwA4FAjrbfx0=" //n
 
 func doBigSegmentsEvaluateSegment(t *ldtest.T) {
 	otherContext := ldcontext.New("other-user-key")
+	otherKind := ldcontext.Kind("other")
 
 	basicSegment := ldbuilders.NewSegmentBuilder("segment1").Version(1).
 		Included(otherContext.Key()). // for "regular included list is ignored for big segment" test
 		Unbounded(true).Generation(100).Build()
-	basicSegmentRef := fmt.Sprintf("%s.g%d", basicSegment.Key, basicSegment.Generation.IntValue())
 
 	segmentWithRule := ldbuilders.NewSegmentBuilder("segment2").Version(1).
 		Unbounded(true).Generation(100).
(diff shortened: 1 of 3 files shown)


sdktests/server_side_big_segments.go
```diff
@@ -27,108 +27,114 @@ var bigSegmentsExpectedHash = "CEXjZY7cHJG/ydFy7q4+YEFwVrG3/pkJwA4FAjrbfx0=" //n
...
 	data := mockld.NewServerSDKDataBuilder().
-		Flag(basicFlag, flagForSegmentWithRule).
-		Segment(basicSegment, segmentWithRule).
+		Flag(basicFlag, flagForSegmentWithRule, flagForSegmentWithOtherKind).
+		Segment(basicSegment, segmentWithRule, segmentWithOtherKind).
 		Build()
 	dataSource := NewSDKDataSource(t, data)
 
 	for _, status := range []ldreason.BigSegmentsStatus{ldreason.BigSegmentsHealthy, ldreason.BigSegmentsStale} {
-		t.Run(fmt.Sprintf("context not found, status %s", status), func(t *ldtest.T) {
```
eli-darkly, line 54: I restructured this block just to change the "test x, status s" subtests into an outer loop for each s plus inner subtests for each x.

sdktests/server_side_big_segments.go
```diff
@@ -139,19 +145,40 @@ func doBigSegmentsEvaluateSegment(t *ldtest.T) {
 		result := evaluateFlagDetail(t, client, basicFlag.Key, otherContext, ldvalue.Null())
 		m.In(t).Assert(result, expectBigSegmentsResult(false, basicFlag, ldreason.BigSegmentsHealthy))
 	})
+
+	t.Run("no query is done if context kind does not match", func(t *ldtest.T) {
```
eli-darkly, line 149: New test.

sdktests/server_side_big_segments.go
```diff
@@ -183,18 +213,56 @@ func doBigSegmentsMembershipCachingTests(t *ldtest.T) {
 		assert.Equal(t, []string{expectedUserHash1}, bigSegmentStore.GetMembershipQueries())
 	})
 
+	t.Run("queries may be done for multiple context kinds", func(t *ldtest.T) {
+		// Here we provide a multi-kind context where there is at least one rule checking each kind,
+		// with a different key for each kind.
```
eli-darkly, line 218: These two tests are new, and cover the biggest change to the big segments evaluation logic: the evaluation engine may now do more than one big segments query, if you have a multi-kind context and it has to evaluate segments that use different kinds (and the keys for those kinds are not the same).

cwaldren-ld (approved):

PR #59, opened 2022-03-31
bucketing tests for the new model

eli-darkly:
This is the v2 equivalent of https://github.com/launchdarkly/sdk-test-harness/pull/56. Besides the basic replacement of users by contexts, the significant differences are:

* There's a new "selection of context" subtest that ensures that whenever we're getting the key or some other attribute based on bucketBy, we are getting it from the right place according to contextKind. That is, if the kind we want is "user", and the context is a single one of kind "user", then we get the attribute directly from that; but if the context is a multi-kind one, then we first look up the "user" kind within it, and then get the attribute. The failure case would be if the SDK just always blindly tried to get the attribute from the top level.
* In the tests involving `bucketBy`, we try both a simple attribute name and a complex attribute reference, to make sure the SDK is really using the attribute reference logic and not just assuming `bucketBy` is a literal attribute name.
* There are some new rules about properties that are not supported in experiments (`bucketBy`, `secondary`). If we're evaluating an experiment, then we just ignore any values that exist for those— so the test is that changing them does not affect the result.
* There's no longer a `NonCritical` loophole for SDKs being inconsistent about whether or not `""` is the same as "undefined" for `secondary`. The new-generation SDKs must all agree that those are _not_ the same.


data/data-files/server-side-eval/rollout-or-experiment.yml
```diff
@@ -0,0 +1,148 @@
+---
+name: rollout or experiment - <NAME>
```
eli-darkly, line 2: This file is the same as the equivalent one in https://github.com/launchdarkly/sdk-test-harness/pull/56 except that I changed the name "user" to "context". The actual context data here is still just simple users, because the point of the test cases here is not to test anything specifically involving the context attributes, but just the general behavior of iterating through the buckets and returning a variation.

data/data-files/server-side-eval/rollout-or-experiment.yml
```diff
@@ -0,0 +1,154 @@
...
+
+# The test cases in this file cover very basic behavior of rollouts and experiments in terms of
+# iterating through the variations and returning one of them. We don't care what the exact bucket
+# value is here-- the weights have been set up so that the tests will work as long as the bucket
+# value is not zero. The detailed tests for the bucket value computation are implemented in Go
+# code in server_side_eval_bucketing.go.
```
eli-darkly, line 8: This file is the same as the equivalent one in https://github.com/launchdarkly/sdk-test-harness/pull/56 except that I changed the name "user" to "context". As the comment says, the specific bucket value doesn't matter for these particular tests, so the context attributes are arbitrary.

sdktests/helpers.go
```diff
@@ -42,13 +48,49 @@ func basicEvaluateFlag(
 	return result.Value
 }
 
+// computeExpectedBucketValue implements the bucketing hash value calculation as per the evaluation spec,
+// except that it returns the value as an integer in the range [0, 100000] - currently the SDKs convert
+// this to a floating-point fraction by in effect dividing it by 100000, but this test code needs an
+// integer in order to compute bucket weights.
+func computeExpectedBucketValue(
```
eli-darkly, line 55: This is identical to the v1 implementation in https://github.com/launchdarkly/sdk-test-harness/pull/56/files#diff-d49f6f508b72b350d5ce430e4f526a838bf90ced9c8a6070cea431494fb446ccR53.

sdktests/server_side_eval_bucketing.go
```diff
@@ -0,0 +1,482 @@
...
+	//
+	// There are also some parameterized tests for rollouts/experiments in the YAML data files. Those cover
+	// more general aspects of the behavior, rather than checking for specific bucket values.
+
+	doTest := func(
+		t *ldtest.T,
+		p bucketingTestParams,
+		rolloutKind ldmodel.RolloutKind,
+		contextKind ldcontext.Kind,
+		bucketBy ldattr.Ref,
```
eli-darkly, line 138: Two differences from the previous implementation: `contextKind` is a new parameter, and `bucketBy` is now a Ref rather than a string.

sdktests/server_side_eval_bucketing.go
```diff
@@ -0,0 +1,482 @@
...
+			doTests(t, makeBucketingTestParams(), ldmodel.RolloutKindRollout, ldcontext.DefaultKind, ldattr.Ref{}, makeContext)
+		})
+
+		t.Run("experiments", func(t *ldtest.T) {
+			doTests(t, makeBucketingTestParamsForExperiments(), ldmodel.RolloutKindExperiment,
+				ldcontext.DefaultKind, ldattr.Ref{}, makeContext)
+		})
+	})
+
+	t.Run("selection of context", func(t *ldtest.T) {
```
eli-darkly, line 271: This test group is new.

sdktests/server_side_eval_bucketing.go
```diff
@@ -0,0 +1,482 @@
...
+				b.Secondary(secondary)
+				if shouldMatchRule {
+					b.SetBool(matchRuleAttr, true)
+				}
+				return b.Build()
+			}
+
+			desc := selectString(secondary == "", "secondary key is an empty string", "secondary key is a non-empty string")
+			t.Run(desc, func(t *ldtest.T) {
+				t.Run("affects bucketing calculations in rollouts", func(t *ldtest.T) {
```
eli-darkly, line 316: Here I removed the `NonCritical` designation in the case where secondary is `""`. U2C-aware SDKs are required to be more consistent in this regard.

sdktests/server_side_eval_bucketing.go
```diff
@@ -0,0 +1,482 @@
...
+							p.seed,
+						)
+						require.NotEqual(t, expectedValueWithoutSecondary, expectedValueWithSecondary)
+						p1 := p
+						p1.overrideExpectedValue = ldvalue.NewOptionalInt(expectedValueWithSecondary)
+						doTest(t, p1, ldmodel.RolloutKindRollout, ldcontext.DefaultKind, ldattr.Ref{}, makeContext)
+					}
+				})
+
+				t.Run("is ignored in experiments", func(t *ldtest.T) {
```
eli-darkly, line 339: New requirement for U2C.

sdktests/server_side_eval_bucketing.go
```diff
@@ -0,0 +1,482 @@
...
+	t.Run("bucket by non-key attribute", func(t *ldtest.T) {
+		t.Run("in rollouts", func(t *ldtest.T) {
+			contextKind := ldcontext.Kind("org")
+
+			t.Run("string value", func(t *ldtest.T) {
+				// For this test group, we'll try it two ways: once with a simple attribute name, and once with a complex
+				// attribute reference. This proves that the SDK is not just assuming bucketBy is a top-level property
+				// name, but is really using the attribute reference logic. We won't bother doing this for the other test
+				// test groups ("integer value", etc.") because the logic for *getting* the attribute is always the same
+				// regardless of what the value of the attribute is.
```
eli-darkly, line 366: Added the "simple attribute vs. complex reference" test variants for U2C.

sdktests/server_side_eval_bucketing.go
```diff
@@ -0,0 +1,482 @@
...
+				for _, p := range makeBucketingTestParams() {
+					p1 := p
+					p1.overrideExpectedValue = ldvalue.NewOptionalInt(0)
+					params = append(params, p1)
+				}
+				doTests(t, params, ldmodel.RolloutKindRollout, ldcontext.DefaultKind, ldattr.NewRef(bucketBy), makeContext)
+			})
+		})
+
+		t.Run("is ignored in experiments", func(t *ldtest.T) {
```
eli-darkly, line 462: New requirement for U2C.

cwaldren-ld (approved):

PR #61, opened 2022-04-01
test for invalid escape sequences in attribute references

eli-darkly:


data/data-files/server-side-eval/errors-bad-attribute-reference.yml
```diff
@@ -0,0 +1,55 @@
...
+name: bad attribute reference errors - <NAME>
+
+parameters:
+  - NAME: clause with no attribute
+    CLAUSE: { "op": "in", "values": [""] }
+    CONTEXT: { kind: "user", key: "user-key" }
+  
+  - NAME: empty path component
+    CLAUSE: { contextKind: "user", "attribute": "///", "op": "in", "values": [""] }
+    CONTEXT: { kind: "user", key: "user-key" }
```
cwaldren-ld, line 11: Could we also have one with `"attribute" : null`, to be sure we treat that the same as no attribute?  

In the Rust SDK, for example, our serialization library sees this differently than omission of the attribute, so we need to be sure to treat it the same.

cwaldren-ld, line 11: Could we also have one with `"attribute" : null`, to be sure we treat that the same as no attribute?  

In the Rust SDK, for example, our serialization library sees this differently than omission of the attribute, so we need to be sure to treat it the same.

cwaldren-ld (approved):

data/data-files/server-side-eval/errors-bad-attribute-reference.yml
eli-darkly: Well... hmm. That one is easy to add, and I thought at first it made sense, but... this is sort of an edge case in our schema that I'm not sure about.

That is: it's possible that in some SDK implementations, explicitly setting this property to null would be considered so invalid that the flag wouldn't even parse, which would break this test. And if so, I'm not sure those SDKs would be wrong. We've sort of made a distinction in our schema between "properties that are optional, and if they're omitted, the default value is ____" (typically, false for booleans, "" for strings)... and "properties that are nullable". An example of the latter would be the `secondary` user attribute, where `null` and `""` are two different things. In those cases we definitely allow an explicit `null`, and treat it the same as omitting the property. But for all the various booleans with a false default, for instance... I think a null is arguably a type mismatch, and that's the kind of thing that (in strongly-typed SDKs at least) would generally make us reject the whole magilla as bad flag data.

cwaldren-ld: Ok, I didn't realize we recognized a distinction there. I will keep this in mind.

eli-darkly: Still thinking about this. One way or the other, we do want to make sure it's clear in specs.

In the meantime I'll merge this PR so we at least have the basic tests.

PR #62, opened 2022-04-07
update goreleaser so we get arm64 binaries

eli-darkly:
We don't have to do anything special in our `.goreleaser.yml` or our publish script to get arm64 builds— those will just automatically include whatever Goreleaser produces by default. We just have to update to a version that actually can produce them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 
-GORELEASER_VERSION=v0.141.0
-GORELEASER_CMD=curl -sL https://git.io/goreleaser | GOPATH=$(mktemp -d) VERSION=$(GORELEASER_VERSION) bash -s -- --rm-dist
+GORELEASER_VERSION=v1.7.0
+GORELEASER_DOWNLOAD_URL=https://github.com/goreleaser/goreleaser/releases/download/v1.7.0/goreleaser_$(shell uname)_$(shell uname -m).tar.gz
+GORELEASER=./bin/goreleaser/goreleaser
 
 GOLANGCI_LINT_VERSION=v1.44.0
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,7 @@
 
-GORELEASER_VERSION=v0.141.0
-GORELEASER_CMD=curl -sL https://git.io/goreleaser | GOPATH=$(mktemp -d) VERSION=$(GORELEASER_VERSION) bash -s -- --rm-dist
```
eli-darkly, line 3: Looks like this script is deprecated— the options for getting a pre-built binary are 1. use a package manager (which would make this less portable between Linux & Mac) or 2. grab it from the GitHub releases using the typical "filename contains platform" pattern.

eli-darkly: [This page](https://goreleaser.com/install/) wrongly claims that you can use `go install` for this purpose— you can certainly get the tool that way, but it would be building it from source.

Makefile
```diff
@@ -24,11 +25,15 @@ test:
 	go test -run=not-a-real-test ./...  # just ensures that the tests compile
 	go test ./...
 
-build-release:
-	$(GORELEASER_CMD) --snapshot --skip-publish --skip-validate
+$(GORELEASER):
```
aengelberg, line 28: I'm not super familiar with `Makefile` syntax; what happens when you use `$()` syntax in the name of a target like this? And why would you want to do that here?

aengelberg (approved):
I had one question about the Makefile syntax, but LGTM :+1:

Makefile
eli-darkly: 1. What happens is just that it substitutes the variable, so the name of this target is really `./bin/goreleaser/goreleaser`.
2. That file won't exist until we download it. So having this target, and having the `build-release` and `publish-targets` depend on this target, is a way of saying "if you don't already have that file, here is how to get it."

eli-darkly: There are 2 basic kinds of makefile targets: ones that represent files, and ones that are targets you would explicitly run like `make build-release`. The latter are the ones that you'll see listed under `.PHONY`, which is a somewhat judgmental term make uses for "this isn't a file, so don't bother checking modification dates or anything like that if there happens to _also_ be a file called `build-release`".

eli-darkly: So, whenever target A has a dependency on target B (written as `A: B`), if B is not a file then make just goes ahead and runs B, but if B a file then make will check whether that file exists/is unchanged since the last build, and only runs the B commands if that is not the case.

aengelberg: Cool! The filename-based makefile target type is definitely the piece I was missing; thanks!

PR #64, opened 2022-04-13
migrate to Go 1.18; use generic optional type & helpers

eli-darkly:
Since this is a standalone codebase, it's a good test case for adopting Go 1.18 in situations where we don't need to support older versions.

The code changes consist of adding two new support packages, `framework/helpers` and `framework/opt`, containing generic-based helper types, and then doing a lot of DRY throughout the project to make use of them:

* `opt.Maybe[V]` replaces `ldvalue.OptionalString`, `ldvalue.OptionalInt`, `ldvalue.OptionalBool`... and all of our uses of pointers to represent optionality, so it's now very hard for us to mess up and cause a nil pointer panic (which I have done more than once in this code).
* The functions in `helpers/channels.go` and related functions replace `expectRequest`, `expectNoRequest`, and many similar cases where we used a select pattern to do a non-blocking receive with timeout or a non-blocking send. They work with channels of any type.
* The functions in `helpers/generics.go` replace `selectString`, `conditionalMatcher`, `stringInSlice`, and similar cases.
* The types in `helpers/test_abstractions.go` help in writing and verifying test helpers. These don't have anything to do with generics, but they were useful in testing the other stuff.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ GORELEASER_VERSION=v1.7.0
 GORELEASER_DOWNLOAD_URL=https://github.com/goreleaser/goreleaser/releases/download/v1.7.0/goreleaser_$(shell uname)_$(shell uname -m).tar.gz
 GORELEASER=./bin/goreleaser/goreleaser
 
-GOLANGCI_LINT_VERSION=v1.44.0
+GOLANGCI_LINT_VERSION=v1.45.2
 
 LINTER=./bin/golangci-lint
 LINTER_VERSION_FILE=./bin/.golangci-lint-version-$(GOLANGCI_LINT_VERSION)
@@ -38,6 +38,7 @@ publish-release: $(GORELEASER)
 $(LINTER_VERSION_FILE):
 	rm -f $(LINTER)
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ./bin $(GOLANGCI_LINT_VERSION)
+	touch $(LINTER_VERSION_FILE)
 
 lint: $(LINTER_VERSION_FILE)
 	$(LINTER) run ./...
--- a/framework/helpers/channels_test.go
+++ b/framework/helpers/channels_test.go
@@ -1,0 +1,78 @@
+package helpers
+
+import (
+	"testing"
+	"time"
+
+	"github.com/launchdarkly/sdk-test-harness/framework/opt"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNonBlockingSend(t *testing.T) {
+	ch1 := make(chan string)
+	assert.False(t, NonBlockingSend(ch1, "a"))
+
+	ch2 := make(chan string, 1)
+	assert.True(t, NonBlockingSend(ch2, "a"))
+	assert.Equal(t, "a", <-ch2)
+}
+
+func TestTryReceive(t *testing.T) {
+	ch := make(chan string, 1)
+	assert.Equal(t, opt.None[string](), TryReceive(ch, time.Millisecond))
+
+	ch <- "a"
+	assert.Equal(t, opt.Some("a"), TryReceive(ch, time.Millisecond))
+
+	go func() {
+		time.Sleep(time.Millisecond * 50)
+		ch <- "b"
+	}()
+	assert.Equal(t, opt.Some("b"), TryReceive(ch, time.Second))
+}
+
+func TestRequireValue(t *testing.T) {
+	tr1 := TestRecorder{PanicOnTerminate: true}
+	ch := make(chan string, 1)
+	assert.PanicsWithValue(t, &tr1, func() { _ = RequireValue(&tr1, ch, time.Millisecond) })
+	if assert.Error(t, tr1.Err()) {
+		assert.Contains(t, tr1.Err().Error(), "waiting for value of type string")
+	}
+
+	tr2 := TestRecorder{PanicOnTerminate: true}
+	ch <- "a"
+	assert.Equal(t, "a", RequireValue(&tr2, ch, time.Millisecond))
+	assert.NoError(t, tr2.Err())
+
+	tr3 := TestRecorder{PanicOnTerminate: true}
+	go func() {
+		time.Sleep(time.Millisecond * 50)
+		ch <- "b"
+	}()
+	assert.Equal(t, "b", RequireValue(&tr3, ch, time.Second))
+	assert.NoError(t, tr3.Err())
+}
+
+func TestRequireNoMoreValues(t *testing.T) {
+	tr1 := TestRecorder{PanicOnTerminate: true}
+	ch := make(chan string, 1)
+	RequireNoMoreValues(&tr1, ch, time.Millisecond)
+	assert.NoError(t, tr1.Err())
+
+	tr2 := TestRecorder{PanicOnTerminate: true}
+	ch <- "a"
+	assert.Panics(t, func() { RequireNoMoreValues(&tr2, ch, time.Millisecond) })
+	if assert.Error(t, tr2.Err()) {
+		assert.Contains(t, tr2.Err().Error(), "extra value of type string")
+	}
+
+	tr3 := TestRecorder{PanicOnTerminate: true}
+	go func() {
+		time.Sleep(time.Millisecond * 50)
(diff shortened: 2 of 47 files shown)


Makefile
```diff
@@ -38,6 +38,7 @@ publish-release: $(GORELEASER)
 $(LINTER_VERSION_FILE):
 	rm -f $(LINTER)
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ./bin $(GOLANGCI_LINT_VERSION)
+	touch $(LINTER_VERSION_FILE)
```
eli-darkly, line 41: This fixes the `make lint` logic so it doesn't re-download the linter unnecessarily.

servicedef/sdk_config.go
```diff
@@ -1,50 +1,51 @@
...
 
 type SDKConfigParams struct {
-	Credential          string                              `json:"credential"`
-	StartWaitTimeMS     ldtime.UnixMillisecondTime          `json:"startWaitTimeMs,omitempty"`
-	InitCanFail         bool                                `json:"initCanFail,omitempty"`
-	Streaming           *SDKConfigStreamingParams           `json:"streaming,omitempty"`
-	Events              *SDKConfigEventParams               `json:"events,omitempty"`
-	PersistentDataStore *SDKConfigPersistentDataStoreParams `json:"persistentDataStore,omitempty"`
-	BigSegments         *SDKConfigBigSegmentsParams         `json:"bigSegments,omitempty"`
-	Tags                *SDKConfigTagsParams                `json:"tags,omitempty"`
```
eli-darkly, line 17: The main advantages of changing these things from pointers to Maybes are 1. it prevents accidental nil panics, and 2. it requires us to use a safe "copy, mutate, copy back" pattern if we want to modify fields within them. The secondary consequence, from the test _service_'s point of view, is that there will be a bunch of explicit nulls if these are unset, instead of just omitting the property, because `json.Marshal`'s `omitempty` behavior _still_ doesn't support dropping zero-valued things that are not of simple types. I left the `omitempty` tags in here just to convey what we would prefer the behavior to be. Maybe one of these days Go will improve it.

testdata/testmodel/eval.go
```diff
@@ -20,10 +21,16 @@ type EvalTestSuite struct {
...
-	User      lduser.User               `json:"user"`
-	ValueType servicedef.ValueType      `json:"valueType"`
-	Default   ldvalue.Value             `json:"default"`
-	Expect    ldreason.EvaluationDetail `json:"expect"`
+	Name      string               `json:"name"`
+	FlagKey   string               `json:"flagKey"`
+	User      lduser.User          `json:"user"`
+	ValueType servicedef.ValueType `json:"valueType"`
+	Default   ldvalue.Value        `json:"default"`
+	Expect    ValueDetail          `json:"expect"`
```
eli-darkly, line 29: I changed the type here from `ldreason.EvaluationDetail` to a very similar new struct type `ValueDetail` so that 1. we don't have interoperability problems between `ldvalue.OptionalInt` (the type of `EvaluationDetail.VariationIndex) and `Maybe[int]`, and 2. on general principle, it's best to avoid relying on implicit behaviors of SDK types (e.g. their JSON property names).

framework/helpers/channels_test.go
```diff
@@ -0,0 +1,78 @@
...
+	assert.False(t, NonBlockingSend(ch1, "a"))
+
+	ch2 := make(chan string, 1)
+	assert.True(t, NonBlockingSend(ch2, "a"))
+	assert.Equal(t, "a", <-ch2)
+}
+
+func TestTryReceive(t *testing.T) {
+	ch := make(chan string, 1)
+	assert.Equal(t, opt.None[string](), TryReceive(ch, time.Millisecond))
```
cwaldren-ld, line 22: Can the compiler infer the generic parameter of string here in `opt.None`? (genuine question - I don't know if this is a feature or not.)

framework/opt/maybe.go
```diff
@@ -0,0 +1,78 @@
...
+// no value if ptr is nil.
+func FromPtr[V any](ptr *V) Maybe[V] {
+	if ptr != nil {
+		return Some[V](*ptr)
+	}
+	return None[V]()
+}
+
+// IsDefined returns true if the Maybe has a value.
+func (m Maybe[V]) IsDefined() bool { return m.defined }
```
cwaldren-ld, line 31: suggestion: since the "defined" variant is called `Some`, this could be `IsSome`.
Could also add `IsNone` for convenience.

sdktests/helpers_test.go
```diff
@@ -38,10 +39,10 @@ func TestComputeExpectedBucketValue(t *testing.T) {
 					p.userValue,
 					p.flagOrSegmentKey,
 					p.salt,
-					ldvalue.NewOptionalString(secondary),
+					o.Some(secondary),
 					p.seed,
 				)
-				failureDesc := selectString(secondary == "", "empty secondary key", "empty-but-not-undefined secondary key") +
+				failureDesc := h.IfElse(secondary == "", "empty secondary key", "empty-but-not-undefined secondary key") +
 					" should have changed result, but did not"
```
cwaldren-ld, line 46: For what it's worth, I do think the `Select` terminology was intuitive 🤷‍♂️

servicedef/service_params.go
```diff
@@ -11,9 +11,8 @@ const (
 	CapabilityAllFlagsClientSideOnly             = "all-flags-client-side-only"
 	CapabilityAllFlagsDetailsOnlyForTrackedFlags = "all-flags-details-only-for-tracked-flags"
 
-	CapabilityPersistentDataStore = "persistent-data-store"
-	CapabilityBigSegments         = "big-segments"
-	CapabilityTags                = "tags"
+	CapabilityBigSegments = "big-segments"
+	CapabilityTags        = "tags"
 )
```
cwaldren-ld, line 16: Q: why was persistent-data-store removed?

cwaldren-ld (approved):

cwaldren-ld:
Nice, I really like these changes.

servicedef/sdk_config.go
eli-darkly: Btw, not surprisingly, the `omitempty` for custom types issue has been raised many times over the years with various reasonable proposals on how to address it, and the Go maintainers have more or less blown off every proposal like "eh, we don't think this is necessary." But their attitude may have changed in the last few years, who knows. I left a comment [here](https://github.com/golang/go/issues/11939#issuecomment-1098521910) about the particular significance of this with generics.

framework/helpers/channels_test.go
eli-darkly: It wouldn't be able to infer it in this particular line no matter what, because the parameter type of `assert.Equal` is just `interface{}` (i.e. there's no compile-time checking that the 2nd & 3rd parameters have the same type - a frequent pitfall in tests that use frameworks like that, and my `matchers` stuff has the same problem, unless & until I rewrite it with generics).

Anyway, the general point was a good question and I didn't know the answer, so I tried and the answer is alas no:

```go

func foo(m opt.Maybe[string]) {}

func bar() {
	foo(opt.None())
	    ^^ compiler error: "cannot infer V"
}
```

framework/opt/maybe.go
eli-darkly: `IsDefined` is what it's called in a lot of optional types I've used on other platforms, and we also used it in the optional types in `ldvalue`. Personally I find it clearer than `IsSome` in a case where you might be looking at some code that is calling `IsSome`, and you don't happen to already know that Some is the name of a constructor. It's also what EasyJSON expects a method like this to be named, even though that's not relevant here.

sdktests/helpers_test.go
eli-darkly: My rationale was that the if { } else { } pattern implies correctly that the first part is what you get if it's true, which the word "select" does not.

servicedef/service_params.go
eli-darkly: The earlier addition of it was a mistake - I haven't implemented that feature yet.

framework/opt/maybe.go
cwaldren-ld: Ah, ok, makes sense. I was coming from the rust perspective where the variant is called Some, and it has IsSome/IsNone.

PR #65, opened 2022-04-14
(#1) update framework to support client-side SDKs, add basic eval tests

eli-darkly:
This is the beginning of client-side contract test support. It includes:

* some refactoring of the underlying framework to allow for more flexible configuration
* some refactoring of the high-level entry points to move the "which test suite is this" logic into `sdktests` instead of `main()`
* changes to the mock streaming service to support the different endpoints and schema used by client-side SDKs
* a new mock polling service (since JS-based client-side SDKs always have polling enabled)
* a parameterized test suite with data files for basic evaluation scenarios - similar to the corresponding ones for server-side, but much simpler because no flag configurations are involved, only preconfigured results; also, for client-side, the user is specified just once when the client is created, not for each evaluation

I've tested this locally with a test service implementation for `node-client-sdk` which I'll put up in its own PR, and the tests pass. I also tried implementing one for `dotnet-client-sdk` to exercise the mobile variant logic, but there are some issues that prevent that one from passing at the moment.


Makefile
```diff
@@ -7,17 +7,21 @@ GOLANGCI_LINT_VERSION=v1.45.2
...
+
+.PHONY: build clean test lint build-release publish-release
 
 build: $(EXECUTABLE)
 
-$(EXECUTABLE): *.go $(shell find . -name '*.go') $(wildcard testdata/data-files/*)
+$(EXECUTABLE): $(SOURCE_FILES) $(TEST_DATA_FILES)
 	go build
 
-pack-files:
```
eli-darkly, line 19: This was never used— I started to add it when I thought I would need to use Packr for the data files.

framework/harness/mock_endpoints.go
```diff
@@ -47,6 +47,34 @@ type MockEndpoint struct {
 	closing     sync.Once
 }
 
+type MockEndpointOption helpers.ConfigOption[MockEndpoint]
+
+type mockEndpointOptionContextFn struct {
+	contextFn func(context.Context) context.Context
```
eli-darkly, line 54: This one isn't new functionality, I just moved it from the regular parameters of NewMockEndpoint to the varargs options. Also, sdk-test-harness doesn't actually use this; it's used by sse-test-harness which shares this framework stuff.

framework/harness/mock_endpoints.go
```diff
@@ -47,6 +47,34 @@ type MockEndpoint struct {
...
+	m.contextFn = o.contextFn
+	return nil
+}
+
+func MockEndpointContextFn(fn func(context.Context) context.Context) MockEndpointOption {
+	return mockEndpointOptionContextFn{fn}
+}
+
+type mockEndpointOptionDescription struct {
+	description string
```
eli-darkly, line 71: There was already a description field in MockEndpoint, but I was never setting it. I've added logic to set it for the various components so we'll see things like "streaming service" in the request logging.

framework/harness/mock_endpoints.go
```diff
@@ -154,16 +182,34 @@ func (m *mockEndpointsManager) serveHTTP(w http.ResponseWriter, r *http.Request)
...
+	wrappedWriter := wrappedResponseWriter{w: w}
+	e.handler.ServeHTTP(&wrappedWriter, transformedReq)
+
+	switch wrappedWriter.status {
+	case http.StatusNotFound:
+		e.logger.Printf("Endpoint %q (%s) received %s request for unrecognized path %s", e.description, e.basePath,
+			r.Method, path)
+	case http.StatusMethodNotAllowed:
+		e.logger.Printf("Endpoint %q (%s) received request with unsupported %s method for path %s", e.description,
+			e.basePath, r.Method, path)
```
eli-darkly, line 217: Just a little extra diagnostic logging for cases where the SDK may have done some kind of incorrect request.

framework/ldtest/filter.go
```diff
@@ -2,12 +2,25 @@ package ldtest
...
 	"regexp"
 	"strings"
 )
 
-// Filter is a function that can determine whether to run a specific test or not.
-type Filter func(TestID) bool
+// Filter is an object that can determine whether to run a specific test or not.
+type Filter interface {
+	Match(id TestID) bool
+}
```
eli-darkly, line 13: This doesn't have much to do with client-side testing, it's just that I realized this API was a bit inconvenient when I started refactoring the top-level entry points.

framework/ldtest/filter.go
```diff
@@ -115,27 +128,13 @@ func PrintFilterDescription(filters RegexFilters, allCapabilities []string, supp
...
-			fmt.Printf("  %s\n", strings.Join(missingCapabilities, ", "))
-			fmt.Println()
+			fmt.Fprintln(out, "Some tests may be skipped because the test service does not support the following capabilities:")
+			fmt.Fprintf(out, "  %s\n", strings.Join(missingCapabilities, ", "))
+			fmt.Fprintln(out)
 		}
 	}
 }
 
-func BriefFilterDescription(filters RegexFilters) string {
```
eli-darkly, line 125: This wasn't ever used.

framework/opt/maybe.go
```diff
@@ -49,6 +50,21 @@ func (m Maybe[V]) OrElse(valueIfUndefined V) V {
 	return valueIfUndefined
 }
 
+// String returns a string representation of the value, or "[none]" if undefined. The string
+// representation of a value is either its own String() if it has such a method, or else the
+// result of fmt.Sprintf with "%v".
```
eli-darkly, line 55: This is important because `String()` gets called for values in test failure messages.

main.go
```diff
@@ -82,22 +80,7 @@ func run(params commandParams) (*ldtest.Results, error) {
...
-	case enabledCapabilities.Has(servicedef.CapabilityClientSide):
-		return nil, errors.New("client-side SDK tests are not yet implemented")
-	default:
-		return nil, errors.New(`test service has neither "client-side" nor "server-side" capability`)
-	}
-
-	fmt.Println()
-	ldtest.PrintFilterDescription(params.filters, allCapabilities, enabledCapabilities)
-
-	results := sdktests.RunServerSideTestSuite(harness, params.filters.Match, testLogger)
```
eli-darkly, line 100: All this logic was moved into `sdktests.RunSDKTestSuite`.

mockld/polling_service.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/launchdarkly/sdk-test-harness/framework"
+
+	"github.com/gorilla/mux"
+)
+
+type PollingService struct {
+	sdkKind     SDKKind
+	currentData SDKData
+	currentEtag string
```
eli-darkly, line 15: I haven't done any Etag stuff in tests yet but this component supports it.

sdktests/client_side_eval.go
```diff
@@ -0,0 +1,175 @@
...
+// or polling behavior-- those will be covered specifically in other tests-- so we just set up the
+// data source to provide one initial data set.
+
+func doClientSideEvalTests(t *ldtest.T) {
+	t.Run("parameterized", runParameterizedClientSideEvalTests)
+}
+
+func runParameterizedClientSideEvalTests(t *ldtest.T) {
+	allTestSuites := getAllClientSideEvalTestSuites(t, "client-side-eval")
+	for _, withReasons := range []bool{false, true} {
```
eli-darkly, line 33: For client-side SDKs, you have to tell the SDK at initialization time whether we'll be using evaluation reasons or not. The main effect of that is the client will add a `withReasons` query string parameter to its requests, which we're not actually checking for here— we're configuring the polling service to return the expected data based on an assumption of what the parameter is. But I'm just being thorough in case for some reason there's some other unexpected difference in SDK behavior depending on the parameter.

sdktests/testapi_sdk_data.go
```diff
@@ -13,44 +13,84 @@ import (
 // SDKDataSource is a test fixture that provides a callback endpoint for SDK clients to connect to,
 // simulating the LaunchDarkly streaming or polling service.
 type SDKDataSource struct {
```
eli-darkly, line 15: I could have made separate SDKStreamingDataSource and SDKPollingDataSource types here, but the exact nature of the data source isn't super relevant in most of the tests— only when we're specifically testing streaming or polling behavior— so I felt like it's still nice to represent it more abstractly in this way.

mockld/polling_service.go
eli-darkly: In fact, as I realized later, there's no way to do anything useful regarding Etags now due to SDK limitations: that is, the SDKs all enforce a minimum polling interval of 30 seconds, so we can't do any caching-related tests that would involve more than one poll unless we create some loophole for that.

sdktests/testapi_sdk_client.go
```diff
@@ -121,9 +132,17 @@ func TryNewSDKClient(t *ldtest.T, configurer SDKConfigurer, moreConfigurers ...S
...
+	if !config.Streaming.IsDefined() && !config.Polling.IsDefined() {
+		return errors.New(
+			"neither streaing nor polling was enabled-- did you forget to include the SDKDataSource as a parameter?")
+	}
+	if config.Streaming.IsDefined() && config.Streaming.Value().BaseURI == "" &&
 		(!config.ServiceEndpoints.IsDefined() || config.ServiceEndpoints.Value().Streaming == "") {
-		return errors.New("streaming base URI was not set-- did you forget to include the SDKDataSource as a parameter?")
+		return errors.New("streaming was enabled but base URI was not set")
+	}
+	if config.Polling.IsDefined() && config.Streaming.Value().BaseURI == "" &&
```
louis-launchdarkly, line 143: -> if config.Polling.IsDefined() && config.**Polling**.Value().BaseURI == "" &&

louis-launchdarkly (approved):
This change refactors the Server correctly to Streaming and Polling, added support to Client Side SDKs (Mobile & JS based) as far as I can tell.

mockld/polling_service.go
eli-darkly: It's absolutely possible to build such a test, but because of how SDK configuration works— and the fact that the test service can only use the same APIs that the customer can use (it doesn't have access to any kind of private APIs)— such a test would take at least 30 seconds to run.

eli-darkly: And, that in itself would not be a terribly valuable test, because if there's a mistake in our implementation of polling, the symptom is very unlikely to be "the SDK stores the results from the first poll, but it doesn't store the results from the second poll." I mean, theoretically anything's possible, but the kinds of things we'd be more interested in testing would be things like:

1. If the first poll got a recoverable error, it correctly polls again 30 seconds later
2. 1. If the first poll got an unrecoverable error, initialization fails and it does not poll again 30 seconds later
3. If the first poll succeeds, and the second one gets a recoverable error, it does a 3rd poll 30 seconds later
4. If the first poll succeeds, and the second one gets an recoverable error, it does not do a 3rd poll 30 seconds later
5. If the second poll returns a new Etag, that Etag value is used for the 3rd request 30 seconds later

1 & 2 would take at least 30 seconds; 3, 4, & 5 would take at least _60_ seconds.

cwaldren-ld (approved):

PR #66, opened 2022-04-14
add contract test for disabling events

eli-darkly:
This is opt-in via a capability, because not all SDKs have a way to say "here is a custom base URI for the events service, but _also_, events are disabled". And we need to be able to set that custom base URI, because otherwise we wouldn't be able to see any events that might wrongly be sent.

### DIFF
--- a/docs/service_spec.md
+++ b/docs/service_spec.md
@@ -47,6 +47,14 @@ This means that the SDK supports Big Segments and can be configured with a custo
 
 For tests that involve Big Segments, the test harness will provide parameters in the `bigSegments` property of the configuration object, including a `callbackUri` that points to one of the test harness's callback services (see [Callback endpoints](#callback-endpoints)). The test service should configure the SDK with its own implementation of a Big Segment store, where every method of the store delegates to a corresponding endpoint in the callback service.
 
+#### Capability `"service-endpoints"`
+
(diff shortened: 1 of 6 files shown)


docs/service_spec.md
```diff
@@ -47,6 +47,14 @@ This means that the SDK supports Big Segments and can be configured with a custo
 
 For tests that involve Big Segments, the test harness will provide parameters in the `bigSegments` property of the configuration object, including a `callbackUri` that points to one of the test harness's callback services (see [Callback endpoints](#callback-endpoints)). The test service should configure the SDK with its own implementation of a Big Segment store, where every method of the store delegates to a corresponding endpoint in the callback service.
 
+#### Capability `"service-endpoints"`
+
+This means that the SDK supports setting the base URIs for the streaming, polling, and events services separately from whether those services are enabled.
```
eli-darkly, line 52: Note that this doesn't require for the SDK API to have a thing that is actually called "service endpoints". All that matters is that the base URI properties/setters are independent of whether the service is enabled.

So for instance, this is true in the C SDKs because there is an independent `LDConfigSetEventsURI` function. And it's true in most of the other SDKs, but in some cases (like Java) when we started moving toward component-scoped configuration with nested builders, we ended up making the base URI properties only available when the service is enabled. We are now walking back from that because it made things a bit inconvenient for people who use Relay Proxy and/or a private LD instance.

cwaldren-ld (approved):

PR #68, opened 2022-04-14
(#2) factor out common eval test logic for client vs. server-side

eli-darkly:
While there's plenty of server-side vs. client-side behavior that's very different, and many tests don't apply to client-side at all, there's also a fair amount of reusable logic where only one or two little pieces are different. Here I've factored that stuff out. That should make the tests easier to maintain (i.e. if we need to improve something in both versions, we can fix it in one place and can't forget to keep them in sync), and also easier to read (because you can see exactly which pieces are different between them, instead of trying to diff them in your head).

I verified that tests still pass against go-server-sdk and node-client-sdk.


framework/helpers/channels.go
```diff
@@ -33,6 +33,7 @@ func TryReceive[V any](ch <-chan V, timeout time.Duration) opt.Maybe[V] {
 // RequireValue tries to receive a value and returns it if successful, or causes the test
 // to fail and terminate immediately if it timed out.
 func RequireValue[V any](t TestContext, ch <-chan V, timeout time.Duration) V {
+	t.Helper()
```
eli-darkly, line 36: Doing this just improves the stacktraces for failures, so the failure line points to whoever called this method.

mockld/polling_service.go
```diff
@@ -36,7 +36,7 @@ func NewPollingService(
 		router.HandleFunc("/sdk/flags/latest-all", pollHandler).Methods("GET")
 	case MobileSDK:
 		router.HandleFunc("/msdk/evalx/users/{user}", pollHandler).Methods("GET")
-		router.HandleFunc("/msdx/evalx/user", pollHandler).Methods("REPORT")
+		router.HandleFunc("/msdk/evalx/user", pollHandler).Methods("REPORT")
```
eli-darkly, line 39: Whoops

eli-darkly: This didn't cause a failure because there isn't currently a test that uses polling mode for mobile.

mockld/polling_service_test.go
```diff
@@ -0,0 +1,89 @@
+package mockld
+
```
eli-darkly, line 2: The polling component I added earlier didn't have unit tests.

cwaldren-ld (approved):

PR #71, opened 2022-04-21
client-side contract tests for tags

eli-darkly:
This includes refactoring the server-side version to share code, similar to what I did with the event tests.


sdktests/common_tests_tags.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/launchdarkly/sdk-test-harness/framework/ldtest"
+	o "github.com/launchdarkly/sdk-test-harness/framework/opt"
+	"github.com/launchdarkly/sdk-test-harness/servicedef"
+
+	"github.com/launchdarkly/go-test-helpers/v2/jsonhelpers"
+	"gopkg.in/launchdarkly/go-sdk-common.v2/lduser"
+
+	"github.com/stretchr/testify/assert"
+)
+
```
eli-darkly, line 18: This file is very similar to what was previously in `server_side_tags.go`. The only relevant differences between client-side and server-side are:

1. client-side SDKs always require one extra config option (the initial user)
2. client-side SDKs in streaming mode sometimes need to also have a polling data source available

eli-darkly: Also, since the server-side SDK test services don't currently support polling mode, I enabled the polling test only for client-side.

cwaldren-ld (approved):

PR #77, opened 2022-04-29
add some client-side streaming tests + refactor common logic

eli-darkly:
This adds tests for streaming behavior of client-side SDKs, in terms of the request method, headers, URL path, and the sending of user properties (either in base64 encoding in the URL for GET, or in the request body for REPORT).

It doesn't cover retry behavior or other streaming tests that are implemented for server-side.

I also did some refactoring of the "common" test stuff in general, since it had become apparent that some of it is pretty repetitive.


mockld/streaming_service.go
```diff
@@ -123,29 +129,66 @@ func (s *StreamingService) makePutEvent() eventsource.Event {
 	}
 }
 
+// Sends on SSE event to all clients that are currently connected to the stream-- or, if no client
+// has connected yet, queues the event so that it will be sent (after the initial data) to the
+// first client that connects. (The latter is necessary to avoid race conditions, since even after
+// a connection is received on the stream endpoint, it is hard for the test logic to know when the
+// HTTP handler has actually created a stream subscription for that connection. If we called
+// the eventsource Publish method before a subscription existed, the event would be lost.)
```
eli-darkly, line 137: The logic here depends on implementation details of `eventsource.Server`, since that's what we're using to actually publish SSE streams. What actually happens when a stream request comes in from the SDK is:

1. The test harness's router figures out what MockEndpoint it corresponds to.
2. It records that we have received a connection.
3. It calls the `eventsource.Server`'s HTTP handler.
4. That handler calls our `ReplayAll` method to provide any _initial_ events for the stream.
5. Then the handler it leaves the connection open, and enters a loop where it sends any events that we choose to publish after that point.

Without the extra synchronization that I added here, it was easy to end up publishing the event after 2 but before 5.

mockld/polling_service.go
```diff
@@ -40,7 +40,7 @@ func NewPollingService(
 		// Note that we only support the "evalx", not the older "eval" which is used only by old unsupported SDKs
 	case JSClientSDK:
 		router.HandleFunc("/sdk/evalx/{env}/users/{user}", pollHandler).Methods("GET")
-		router.HandleFunc("/sdk/evalx/{env}/users", pollHandler).Methods("REPORT")
+		router.HandleFunc("/sdk/evalx/{env}/user", pollHandler).Methods("REPORT")
```
eli-darkly, line 43: The actual URL that the SDKs use here does not have a trailing `s`.

sdktests/common_tests_stream_request.go
```diff
@@ -0,0 +1,162 @@
...
+
+				if method == flagRequestREPORT {
+					m.In(t).For("request body").Assert(request.Body, m.AllOf(
+						m.Not(m.BeNil()),
+						userJSONMatcher))
+				} else {
+					m.In(t).For("request body").Assert(request.Body, m.Length().Should(m.Equal(0)))
+
+					pathParts := strings.Split(strings.TrimPrefix(request.URL.Path, "/"), "/")
+					expectedPathComponents := h.IfElse(c.sdkKind == mockld.JSClientSDK, 3, 2)
```
eli-darkly, line 150: This logic regarding the path is a bit clunky - I've improved it in the next PR.

sdktests/testsuite_entry_point.go
```diff
@@ -70,7 +70,6 @@ func RunSDKTestSuite(
 }
 
 func doAllServerSideTests(t *ldtest.T) {
-	t.Run("data store", doServerSideDataStoreTests)
 	t.Run("evaluation", doServerSideEvalTests)
```
cwaldren-ld, line 73: I think I missed this - so have these tests been moved into one of the common files?

eli-darkly: Sorry, I forgot to comment on that: they are moved and renamed, so they're now the "updates" part of the "stream" tests. They are really testing both the stream update functionality _and_ the update-versioning functionality that's part of the in-memory data store, but there's not really a way to test the latter without the former, and any failures are much more likely to be from the former, so I figured it was clearer to call them stream tests.

eli-darkly: See:
* https://github.com/launchdarkly/sdk-test-harness/pull/77/files#diff-c366be2f3b60d321e72c92cf2eb34a673ee62965ee6eb6293146bcec3a671515R11
* https://github.com/launchdarkly/sdk-test-harness/pull/77/files#diff-3fd8a54b081125b9520ab25675cb065e143000abd9da7131129a99e25fc4507fR13
* https://github.com/launchdarkly/sdk-test-harness/pull/77/files#diff-6b419db42cb6d705a3cd03fc68e8e5f209dcd3aa944d48edb61ede2a518e7cbb

cwaldren-ld (approved):

PR #79, opened 2022-04-29
allow api_key scheme in auth header

eli-darkly:
I categorized this as a client-side test fix just because currently (I think) the mobile SDKs are the only ones that like to send `Authorization: api_key xxx` rather than just `Authorization: xxx`. Including the `api_key` scheme identifier is technically a correct thing to do (and probably better than omitting it), and we haven't been consistent about it, and so the tests need to allow both (as LD does).


keelerm84 (approved):
Thank you for fixing this so quickly

PR #83, opened 2022-05-19
miscellaneous fixes for client-side U2C behavior

eli-darkly:
The initial implementation of U2C tests was for server-side. Client-side tests had been merged forward from the main branch, but hadn't yet been tested against a client-side U2C SDK, and there were various problems that showed up once I started testing against an in-progress dotnet-client-sdk.

### DIFF
--- a/sdktests/common_tests_eval.go
+++ b/sdktests/common_tests_eval.go
@@ -78,7 +78,9 @@ func (c CommonEvalParameterizedTestRunner[SDKDataType]) runTestSuite(
 
 	// We can't rely on the test framework's usual auto-closing of the client, because this method could
 	// be called multiple times for a single value of t.
-	defer client.Close()
+	defer func() {
+		_ = client.Close()
+	}()
 
 	if len(suite.Evaluations) == 1 && suite.Evaluations[0].Name == "" {
 		c.runTestEval(t, suite, suite.Evaluations[0], sdkData, client)
--- a/sdktests/common_tests_eval.go
+++ b/sdktests/common_tests_eval.go
@@ -60,7 +60,6 @@ func (c CommonEvalParameterizedTestRunner[SDKDataType]) runTestSuite(
 	t *ldtest.T,
 	suite testmodel.EvalTestSuite[SDKDataType],
 ) {
-	t.Debug("*** yo %+v", suite)
 	if suite.RequireCapability != "" {
 		t.RequireCapability(suite.RequireCapability)
 	}
@@ -76,6 +75,10 @@ func (c CommonEvalParameterizedTestRunner[SDKDataType]) runTestSuite(
 		clientConfig = c.SDKConfigurers(suite)
 	}
 	client := NewSDKClient(t, append(clientConfig, dataSource)...)
+
+	// We can't rely on the test framework's usual auto-closing of the client, because this method could
+	// be called multiple times for a single value of t.
+	defer client.Close()
 
 	if len(suite.Evaluations) == 1 && suite.Evaluations[0].Name == "" {
 		c.runTestEval(t, suite, suite.Evaluations[0], sdkData, client)
--- a/sdktests/common_tests_events_contexts.go
+++ b/sdktests/common_tests_events_contexts.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/launchdarkly/sdk-test-harness/v2/data"
+	h "github.com/launchdarkly/sdk-test-harness/v2/framework/helpers"
 	"github.com/launchdarkly/sdk-test-harness/v2/framework/ldtest"
 	o "github.com/launchdarkly/sdk-test-harness/v2/framework/opt"
 	"github.com/launchdarkly/sdk-test-harness/v2/mockld"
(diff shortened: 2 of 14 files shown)


sdktests/common_tests_eval.go
```diff
@@ -77,6 +76,12 @@ func (c CommonEvalParameterizedTestRunner[SDKDataType]) runTestSuite(
 	}
 	client := NewSDKClient(t, append(clientConfig, dataSource)...)
 
+	// We can't rely on the test framework's usual auto-closing of the client, because this method could
+	// be called multiple times for a single value of t.
+	defer func() {
+		_ = client.Close()
+	}()
```
eli-darkly, line 83: Neglecting to close an SDK client before starting another SDK client is not a problem for server-side SDKs, and it's not a problem with node-client-sdk which I had been testing with earlier, but it does cause a failure in SDKs like dotnet-client-sdk that enforce singleton usage. It's not an issue in most of the tests because any client created within the scope of a `t *ldtest.T` gets automatically closed at the end of that test or subtest— but in this particular case, this helper method wasn't always associated with a single test scope.

sdktests/common_tests_events_contexts.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/launchdarkly/sdk-test-harness/v2/data"
```
eli-darkly, line 6: Most of the changes in this file aren't related to client-side usage (see next comment for one that is), but I revised the logic here for better code sharing. Previously, in each of these test cases I had to construct a matcher that exactly described the whole JSON object we were looking for. I wasn't able to use the `JSONMatchesContext` helper to do that for me, because it only handled the regular schema, not the event schema. But I refactored that helper so now it can do both. As a result, instead of having to describe the whole expected JSON object, we just have to make any expected changes to the context (like removing redacted attributes) and say which redacted attribute names we expect to see, if any.

sdktests/common_tests_events_contexts.go
```diff
@@ -169,23 +145,26 @@ func (c CommonEventTests) EventContexts(t *ldtest.T) {
...
+				expectedContext = p.outputContext(context)
+			}
+			return JSONMatchesEventContext(expectedContext, p.redactedShouldBe)
+		}
+
 		t.Run(p.name, func(t *ldtest.T) {
 			contexts := p.contextFactory("doServerSideEventContextTests")
 			events := NewSDKEventSink(t)
-			client := NewSDKClient(t, WithEventsConfig(p.eventsConfig), dataSource, events)
+			client := NewSDKClient(t, c.baseSDKConfigurationPlus(WithEventsConfig(p.eventsConfig), dataSource, events)...)
```
eli-darkly, line 159: Omitting this caused it to initialize the client-side SDK test service without the required initial user parameter.

sdktests/common_tests_events_contexts.go
```diff
@@ -169,23 +145,26 @@ func (c CommonEventTests) EventContexts(t *ldtest.T) {
...
-					return matchers
-				}
-				return append([]m.Matcher{IsIndexEvent()}, matchers...)
-			}
-
 			t.Run("debug event", func(t *ldtest.T) {
 				context := contexts.NextUniqueContext()
+				if c.isClientSide {
+					client.SendIdentifyEvent(t, context)
+				}
```
eli-darkly, line 167: For a client-side SDK we have to set the current user before doing the evaluation.

sdktests/common_tests_events_custom.go
```diff
@@ -57,14 +57,16 @@ func (c CommonEventTests) CustomEvents(t *ldtest.T) {
 				client.FlushEvents(t)
 				payload := events.ExpectAnalyticsEvents(t, defaultEventTimeout)
 
-				m.In(t).Assert(payload, m.ItemsInAnyOrder(
-					IsIndexEvent(),
```
eli-darkly, line 61: Client-side SDKs won't send an index event.

sdktests/custom_matchers_contexts.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
+	m "github.com/launchdarkly/go-test-helpers/v2/matchers"
+)
+
+// JSONMatchesContext builds a Matcher to verify that the input JSON is a valid representation of the
+// specified context. This is using the regular context schema (i.e. what would be sent to evaluation
+// endpoints), not the event schema.
+//
+// The matcher should be tolerant of all allowable variants: for instance, it is legal to include
+// `"transient": false` in the representation rather than omitting transient.
```
eli-darkly, line 16: The logic for this is a bit convoluted, but I think it's worthwhile so that we don't have to embed logic like this throughout various tests. There are unit tests in the next file to verify the behavior, so as long as those are covering enough valid and invalid input combinations, we should be able to rely on this matcher.

sdktests/testapi_sdk_client.go
```diff
@@ -107,7 +110,16 @@ type SDKClient struct {
 // The object's lifecycle is tied to the test scope that created it; it will be automatically closed
 // when this test scope exits. It can be reused by subtests until then.
 func NewSDKClient(t *ldtest.T, configurers ...SDKConfigurer) *SDKClient {
+	count := atomic.AddInt32(&currentlyExistingClientInstances, 1)
+	if count == 2 && t.Capabilities().Has(servicedef.CapabilitySingleton) {
+		atomic.AddInt32(&currentlyExistingClientInstances, -1)
+		t.Errorf("Test tried to create an SDK client instance when one already existed, and this SDK is a singleton")
```
eli-darkly, line 116: The test service would have failed anyway when we tried to create another client, if it doesn't support multiple clients. But failing here is faster.

PR #84, opened 2022-06-16
Run "context type" tests against client SDKs

keelerm84:


eli-darkly (commented):
You know... I think your earlier suggestion of "we should add a default initial context if we didn't provide one" is better. I'll put in a PR for that and then I think some things will be simpler. My original concern was that it would allow us to forget to set one in tests where it does matter, but on second thought I can't really think of a scenario where that would be an issue.

keelerm84:
> I think some things will be simpler.

Just a smidge. :smile: 

Thanks for making that other PR!

PR #88, opened 2022-07-11
use our own implementation of Eventually/Never assertions to avoid panics

eli-darkly:
This is a more reliable fix for https://github.com/launchdarkly/sdk-test-harness/pull/29. It's desirable for us to have a way to force early test termination from within any helper function, regardless of where that function is being called from; doing so from within an `assert.Eventually` type of polling loop shouldn't inherently be bad. The problem is just that the actual implementation of `assert.Eventually` uses a separate goroutine— which has the benefit of making the function work as intended even if the function being polled just hangs or something, but it is not compatible with the idea of early termination. So, here I've implemented a simpler version of that logic and we're fine as long as we use that.

### DIFF
--- a/framework/helpers/assertions.go
+++ b/framework/helpers/assertions.go
@@ -1,0 +1,100 @@
+package helpers
+
+import (
+	"time"
+)
+
+// Calls testFn repeatedly at intervals until the expected value is seen or the timeout elapses.
+// Returns true if the value was matched, false if timed out.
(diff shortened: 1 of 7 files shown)


aengelberg (commented):
I'm about to look through the code now so I can give my approval, but I can confirm that this fixes my problem, as I tried running it locally against my test service and it works.

aengelberg (approved):
Thanks, looks good!

PR #89, opened 2022-07-11
allow extra "device" & "os" attributes for mobile only

eli-darkly:
This fixes the user matching logic to work correctly for mobile SDKs that add "device" and "os" custom attributes. It allows these attributes to have any values, and does not _require_ mobile SDKs to add them, but it does not allow any non-mobile SDKs to add them.

The logic for this is a bit convoluted, but it goes away in v2 because we won't be adding these attributes any more in upcoming major versions of the mobile SDKs.

### DIFF
--- a/framework/helpers/generics.go
+++ b/framework/helpers/generics.go
@@ -1,5 +1,16 @@
 package helpers
 
+import (
+	"sort"
+
+	"golang.org/x/exp/constraints"
+)
+
+// CopyOf returns a shallow copy of a slice.
+func CopyOf[V any](slice []V) []V {
+	return append([]V(nil), slice...)
+}
+
 // IfElse returns valueIfTrue or valueIfFalse depending on isTrue.
 func IfElse[V any](isTrue bool, valueIfTrue, valueIfFalse V) V {
 	if isTrue {
@@ -16,4 +27,11 @@ func SliceContains[V comparable](value V, slice []V) bool {
 		}
 	}
 	return false
+}
+
+// Sorted returns a sorted copy of a slice.
+func Sorted[V constraints.Ordered](slice []V) []V {
+	ret := CopyOf(slice)
+	sort.Slice(ret, func(i, j int) bool { return ret[i] < ret[j] })
+	return ret
 }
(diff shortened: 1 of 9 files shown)


aengelberg (commented):
I just tested this locally, and it looks like some tests are still failing, because if `allAttributesPrivate=true`, that includes the `device` and `os` attributes. So the privateAttrs array isn't matching properly, because it includes those two fields sometimes:
```
[events/user properties/inlineUsers=true, allAttributesPrivate=true/user-private=none/identify event]
  item[0] JSON property "user" JSON property "privateAttrs" in order did not equal ["country", "firstName", "lastName", "preferredLanguage", "primaryLanguage"], Unexpected JSON property key(s) [anonymous]; allowed keys are [custom, key, privateAttrs]
  full value was: [{"creationDate":1657582933685,"key":"doClientSideEventUserTests.1657582928310.1657582928310.37","kind":"identify","user":{"anonymous":false,"key":"doClientSideEventUserTests.1657582928310.1657582928310.37","privateAttrs":["firstName","device","country","lastName","primaryLanguage","os","preferredLanguage"]}}]
  Stacktrace:
    sdktests.CommonEventTests.eventUsersWithConfig.func3.1 (common_tests_events_users.go:90)
    sdktests.CommonEventTests.eventUsersWithConfig.func3 (common_tests_events_users.go:84)
    sdktests.CommonEventTests.eventUsersWithConfig (common_tests_events_users.go:83)
    sdktests.CommonEventTests.EventUsers.func2 (common_tests_events_users.go:32)
    sdktests.CommonEventTests.EventUsers (common_tests_events_users.go:31)
    sdktests.doClientSideEventUserTests (client_side_events_all.go:63)
    sdktests.doClientSideEventTests (client_side_events_all.go:17)
    sdktests.doAllClientSideTests (testsuite_entry_point.go:84)
    sdktests.RunSDKTestSuite.func1 (testsuite_entry_point.go:67)
```

aengelberg (approved):
Ran it locally, can confirm it does seem to address all device / os related failures.

PR #90, opened 2022-07-11
Ignore blank lines in external suppression file

aengelberg:
When keeping a list of test suppressions for a given SDK, I have been reaching for comments and blank lines to make the file readable and informative, similarly to a `.gitignore`:
```sh
# bug in SDK: see ticket sc-12345
test/suppression/x
test/suppression/y

# another bug: see ticket sc-67890
test/suppression/z
```
Comments don't need special treatment for this to work, because the test harness will read in the commented lines as test names and match nothing. But blank lines currently make the test harness skip _everything_, because it thinks I'm skipping `""` which matches every test.

I'd like to add in the ability to skip over blank lines here, to enable that use case.

### DIFF
--- a/main.go
+++ b/main.go
@@ -118,7 +118,12 @@ func loadSuppressions(params *commandParams) error {
 	defer func() { _ = file.Close() }()
 	scanner := bufio.NewScanner(file)
 	for scanner.Scan() {
-		escaped := regexp.QuoteMeta(scanner.Text())
+		line := scanner.Text()
+		// Ignore blank lines
+		if strings.TrimSpace(line) == "" {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


eli-darkly:
I would prefer to explicitly skip comments too, rather than rely on them not matching anything. For one thing, the test output includes some stuff about "some tests will be skipped based on the filter parameters, which were such-and-such" and it's pointless to have those include garbage from comments.

eli-darkly (approved):
This has been left in a draft state for a while - not sure if you were going to revisit it. I do think it would be a good idea to explicitly skip comment lines as I suggested, but I'd be fine with you merging it like this too.

PR #97, opened 2022-08-02
Add kind and array based matching operation checks

keelerm84:
There was an issue in the rust eval crate (sc-162733) where the match
operator was evaluating regexes incorrectly if:

1. It was testing against the context's kind, or
2. It was testing against an array of context values

This commit should add tests that cover both of the scenarios.


eli-darkly (approved):
Looks good - I have a slightly different idea of how to organize these but I'm not sure my idea makes sense yet, so I'll fiddle with it as a follow-up but in the meantime we should merge this.

PR #103, opened 2022-09-19
Scaffold location for open feature tests.

kinyoklion:
Drafting this for some questions.

I want to determine if this approach is favorable.

There are broadly a few ways to implement open feature tests.
- Implement them as a sdkKind similar to client/server. I think this makes the most sense from an external perspective. It is a different kind of thing. It is based on a server SDK though. What this PR does.
- Implement a capability that would exercise the open feature methods and could be ran in combination with other server tests.

Independent of this there are a couple ways to handle evaluation parameters/results. 
- Implement specific types for the parameters and result. The method I have taken so far. There is some duplicate work.
- Implement a translation that provides the results in a way compatible with the existing evaluation. Benefit being we could run single evaluations (not all flags) that are specified in existing suites. The big down side is most of OpenFeature work is converting these types, so it makes it hard to know if that is working completely correctly.

### DIFF
--- a/mockld/streaming_service.go
+++ b/mockld/streaming_service.go
@@ -78,6 +78,8 @@ func NewStreamingService(
 	streamHandler := streams.Handler(allDataChannel)
 	router := mux.NewRouter()
 	switch sdkKind {
+	case ServerSideOpenFeature:
+		fallthrough
 	case ServerSideSDK:
 		router.HandleFunc(StreamingPathServerSide, streamHandler).Methods("GET")
 	case MobileSDK:
--- a/sdktests/server_side_open_feature_eval.go
+++ b/sdktests/server_side_open_feature_eval.go
@@ -1,6 +1,27 @@
 package sdktests
 
-import "github.com/launchdarkly/sdk-test-harness/framework/ldtest"
+import (
+	"github.com/launchdarkly/sdk-test-harness/framework/ldtest"
+	"github.com/launchdarkly/sdk-test-harness/mockld"
+	"github.com/launchdarkly/sdk-test-harness/servicedef"
+	"gopkg.in/launchdarkly/go-sdk-common.v2/ldvalue"
+	"gopkg.in/launchdarkly/go-server-sdk-evaluation.v1/ldbuilders"
+)
 
 func doServerSideOpenFeatureEvalTests(t *ldtest.T) {
+	expectedValueV1 := ldvalue.Int(1)
+	flagKey := "flag"
+	flag := ldbuilders.NewFlagBuilder(flagKey).Version(1).
+		On(false).OffVariation(0).Variations(expectedValueV1).Build()
+	data := mockld.NewServerSDKDataBuilder().Flag(flag).Build()
+	dataSource := NewSDKDataSource(t, data)
+	client := NewSDKClient(t, dataSource)
+
+	client.EvaluateOpenFeatureFlag(t, servicedef.OpenFeatureEvaluateFlagParams{
+		FlagKey:           "flag",
+		EvaluationContext: map[string]string{"targetingKey": "key"},
+		ValueType:         servicedef.ValueTypeInt,
+		DefaultValue:      ldvalue.Int(0),
+		Detail:            true,
+	})
 }
--- a/sdktests/server_side_open_feature_eval.go
+++ b/sdktests/server_side_open_feature_eval.go
@@ -1,6 +1,7 @@
 package sdktests
 
 import (
+	"fmt"
 	"github.com/launchdarkly/sdk-test-harness/framework/ldtest"
 	"github.com/launchdarkly/sdk-test-harness/mockld"
 	"github.com/launchdarkly/sdk-test-harness/servicedef"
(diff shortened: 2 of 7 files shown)


mockld/streaming_service.go
```diff
@@ -78,6 +78,8 @@ func NewStreamingService(
 	streamHandler := streams.Handler(allDataChannel)
 	router := mux.NewRouter()
 	switch sdkKind {
+	case ServerSideOpenFeature:
+		fallthrough
```
kinyoklion, line 82: Could copy/paste if that is more favorable.

sdktests/server_side_open_feature_eval.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+	"github.com/launchdarkly/sdk-test-harness/framework/ldtest"
+	"github.com/launchdarkly/sdk-test-harness/mockld"
+	"github.com/launchdarkly/sdk-test-harness/servicedef"
+	"gopkg.in/launchdarkly/go-sdk-common.v2/ldvalue"
+	"gopkg.in/launchdarkly/go-server-sdk-evaluation.v1/ldbuilders"
+)
+
+func doServerSideOpenFeatureEvalTests(t *ldtest.T) {
+	expectedValueV1 := ldvalue.Int(1)
```
kinyoklion, line 13: I've only put a test directly here to verify that things happen. This is not an example of an actual setup. I would make the common files, but I wanted to validate the other elements before doing that.

sdktests/testapi_sdk_client.go
```diff
@@ -209,6 +209,22 @@ func (c *SDKClient) EvaluateFlag(t *ldtest.T, params servicedef.EvaluateFlagPara
 	return resp
 }
 
+func (c *SDKClient) EvaluateOpenFeatureFlag(t *ldtest.T, params servicedef.OpenFeatureEvaluateFlagParams) servicedef.OpenFeatureEvaluateFlagResponse {
```
kinyoklion, line 212: Maybe I could make this less duplicated with generics?

kinyoklion:
I just want to get feedback on the approach at the moment. Not a specific test.

eli-darkly:
I have mixed feelings about the basic premise of this— that is, reusing `sdk-test-harness` for OpenFeature testing.

I understand the desire to avoid reinventing the wheel, and the basic concept of contract testing seems like it could be applicable to these libraries.

However... what we would be testing has a different shape. It's a different abstraction, even if it's an LD SDK under the covers. The entire sdk-test-harness protocol is designed for the SDK model— so, for instance, the evaluate command takes a _user_. Would the test service for an OpenFeature library take a user, and then transform that into an OF evaluation context, and pass that to the OF client which would then end up transforming it back into a user? Or would we be adding new commands to the protocol which is _only_ relevant for OpenFeature clients, and saying that if it's an OpenFeature client then we will never send any of the other commands that are for the SDKs, and the test suites will be implemented totally separately?

If so, I'm not sure how much value we would be getting from implementing these similar but incompatible things in the same project. If on the other hand we just wanted to reuse the lower-level concepts, like the basic HTTP mechanism and the `ldtest` framework... those are easily reusable by another project. That's why they have their own parent package, `framework`. My goal with that was to eventually get rid of the equivalent stuff in `sse-test-harness`, and have `sse-test-harness` simply import those packages from `sdk-test-harness`.

kinyoklion:
> I have mixed feelings about the basic premise of this— that is, reusing `sdk-test-harness` for OpenFeature testing.
> 
> I understand the desire to avoid reinventing the wheel, and the basic concept of contract testing seems like it could be applicable to these libraries.
> 
> However... what we would be testing has a different shape. It's a different abstraction, even if it's an LD SDK under the covers. The entire sdk-test-harness protocol is designed for the SDK model— so, for instance, the evaluate command takes a _user_. Would the test service for an OpenFeature library take a user, and then transform that into an OF evaluation context, and pass that to the OF client which would then end up transforming it back into a user? Or would we be adding new commands to the protocol which is _only_ relevant for OpenFeature clients, and saying that if it's an OpenFeature client then we will never send any of the other commands that are for the SDKs, and the test suites will be implemented totally separately? In your example code here, it looks like you're doing the latter.
> 
> If so, I'm not sure how much value we would be getting from implementing these similar but incompatible things in the same project. If on the other hand we just wanted to reuse the lower-level concepts, like the basic HTTP mechanism and the `ldtest` framework... those are easily reusable by another project without copying anything. That's why they have their own parent package, `framework`. My goal with that was to eventually get rid of the equivalent stuff in `sse-test-harness`, and have `sse-test-harness` simply import those packages from `sdk-test-harness`.

So

> I have mixed feelings about the basic premise of this— that is, reusing `sdk-test-harness` for OpenFeature testing.
> 
> I understand the desire to avoid reinventing the wheel, and the basic concept of contract testing seems like it could be applicable to these libraries.
> 
> However... what we would be testing has a different shape. It's a different abstraction, even if it's an LD SDK under the covers. The entire sdk-test-harness protocol is designed for the SDK model— so, for instance, the evaluate command takes a _user_. Would the test service for an OpenFeature library take a user, and then transform that into an OF evaluation context, and pass that to the OF client which would then end up transforming it back into a user? Or would we be adding new commands to the protocol which is _only_ relevant for OpenFeature clients, and saying that if it's an OpenFeature client then we will never send any of the other commands that are for the SDKs, and the test suites will be implemented totally separately? In your example code here, it looks like you're doing the latter.
> 
> If so, I'm not sure how much value we would be getting from implementing these similar but incompatible things in the same project. If on the other hand we just wanted to reuse the lower-level concepts, like the basic HTTP mechanism and the `ldtest` framework... those are easily reusable by another project without copying anything. That's why they have their own parent package, `framework`. My goal with that was to eventually get rid of the equivalent stuff in `sse-test-harness`, and have `sse-test-harness` simply import those packages from `sdk-test-harness`.

I think the biggest thing that is re-used is that all the configuration for the SDK is the same for OpenFeature as it is for an LD SDK. Because we are directly creating the SDK. But that could be shared in some way as well. It isn't a test for any open feature provider implementation, only for LD SDKs.

As the only capability of an open feature client itself is evaluation, I was only focusing on just evaluation. In which case the test takes an evaluation context, not a user, because that is the shape of things for OpenFeature. In the provider though that gets converted to a user. and that could be verified in events.

That said, if you do an evaluation against a context, then the events also should come out the same.

PR #105, opened 2022-10-04
implement evaluation tests for PHP

eli-darkly:
This is the first step in adding contract test support for the PHP SDK. Due to the PHP SDK's atypical requirements and architecture, many of the contract tests aren't applicable to it, but we do want to at least verify its evaluation engine and event behavior; this PR is just for the evaluation tests.

Fortunately, the architecture of sdk-test-harness makes this a not-huge task. The server-side evaluation tests can actually be reused in their entirety; the special behavior for PHP only needs to be implemented in the fake data source component. Basically, each of the evaluation tests is configured with a full set of flag/segment data (like: `{"flags": {"flag1": {...}, "flag2": {...}}, "segments": {...}}`), and if we were testing a regular server-side SDK, the data source endpoint would return literally that data. But if we're testing the PHP SDK, we instead create multiple data source endpoints (`/flags`, `/flags/{key}`, `/segments/{key}`), corresponding to the request paths used by the PHP SDK, and each is configured to extract the appropriate subset of the configured data.

In the development repo for the PHP SDK, there is a WIP branch for the test service implementation to make this work.

### DIFF
--- a/docs/service_spec.md
+++ b/docs/service_spec.md
@@ -21,11 +21,11 @@ This resource should return a 200 status to indicate that the service has starte
 
 The test harness will use the `capabilities` information to decide whether to run optional parts of the test suite that relate to those capabilities.
 
-#### SDK type capabilities: `"server-side"`, `"client-side"`, `"mobile"`
+#### SDK type capabilities: `"server-side"`, `"client-side"`, `"mobile"`, `"php"`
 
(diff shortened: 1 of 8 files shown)


keelerm84 (approved):
That's awesome this was so much easier than expected.

PR #112, opened 2022-10-20
make "context-type" a separate capability

eli-darkly:
Tests that require the existence of a custom type for evaluation contexts need to be gated by a specific capability. It's not the same as the "strongly-typed" capability, which refers to whether _variation methods_ are strongly typed.

This will cause those tests to be temporarily disabled in existing SDKs until those have updated their contract test capabilities.


keelerm84 (commented):
These changes look good. I think there are also a couple of references in the doc to the strongly-typed capability that need updated.

[Here](https://github.com/launchdarkly/sdk-test-harness/blob/eb/sc-171550/more-php-events/docs/service_spec.md#build-a-context) and [here](https://github.com/launchdarkly/sdk-test-harness/blob/eb/sc-171550/more-php-events/docs/service_spec.md#convert-a-context).

cwaldren-ld (approved):

PR #113, opened 2022-10-25
fix client-side SDK event tests that were broken by PHP changes (also bump dependencies)

eli-darkly:
I made a mistake in https://github.com/launchdarkly/sdk-test-harness/pull/110, when I was updating the event tests to allow for PHP: when it's _not_ PHP, we always should expect a summary event after an evaluation. I had mistakenly dropped the summary event expectation for client-side SDKs.

Also, just to keep things generally up to date, I've bumped the versions of prerelease SDK packages, and removed a little test helper code that's no longer relevant due to functionality that was removed from the U2C spec.


sdktests/common_tests_events_contexts.go
```diff
@@ -221,7 +221,7 @@ func (c CommonEventTests) EventContexts(t *ldtest.T) {
 						),
 					}
 					if c.isClientSide {
-						eventMatchers = append(eventMatchers, IsIdentifyEvent())
+						eventMatchers = append(eventMatchers, IsIdentifyEvent(), IsSummaryEvent())
```
eli-darkly, line 224: This is the fix.

sdktests/helpers.go
```diff
@@ -241,21 +241,12 @@ func pollUntilFlagValueUpdated(
...
-			value = arrayBuilder.Build()
-		} else {
-			objectBuilder := ldvalue.ObjectBuild()
-			objectBuilder.Set(name, value)
-			value = objectBuilder.Build()
-		}
+		name := ref.Component(depth)
+		objectBuilder := ldvalue.ObjectBuild()
+		objectBuilder.Set(name, value)
+		value = objectBuilder.Build()
```
eli-darkly, line 247: Adjusted this because the SDK no longer supports array indexing in attribute references.

cwaldren-ld (approved):

PR #116, opened 2022-11-15
add tests for feature & debug events

eli-darkly:
This fills in the last remaining gaps in client-side SDK event test coverage as far as I know, in terms of the types of events being tested.

As with the previous PR, I didn't manage to come up with an abstraction to allow much common test logic to be shared between server-side and client-side tests here— the nature of the flag data is just too different.

### DIFF
--- a/mockld/sdk_data.go
+++ b/mockld/sdk_data.go
@@ -252,6 +252,11 @@ func (b *ClientSDKDataBuilder) Flag(key string, props ClientSDKFlag) *ClientSDKD
 	return b
 }
 
+func (b *ClientSDKDataBuilder) FullFlag(props ClientSDKFlagWithKey) *ClientSDKDataBuilder {
+	b.flags[props.Key] = props.ClientSDKFlag
+	return b
+}
+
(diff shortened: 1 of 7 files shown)


mockld/sdk_data.go
```diff
@@ -252,6 +252,11 @@ func (b *ClientSDKDataBuilder) Flag(key string, props ClientSDKFlag) *ClientSDKD
 	return b
 }
 
+func (b *ClientSDKDataBuilder) FullFlag(props ClientSDKFlagWithKey) *ClientSDKDataBuilder {
+	b.flags[props.Key] = props.ClientSDKFlag
+	return b
+}
+
```
eli-darkly, line 259: The client-side data is a little funny because for flag patches, you want an object that includes a `key` property (ClientSDKFlagWithKey), but for the initial data or stream `put` event, you want it without the `key` (ClientSDKFlag). ClientSDKFlagWithKey is more convenient for generating test data since it includes all of the properties in one struct, so I added this method to let us use it everywhere.

cwaldren-ld (approved):

PR #121, opened 2022-11-30
fix several test problems for client-side SDKs

eli-darkly:
See PR comments.

### DIFF
--- a/sdktests/common_tests_events_contexts.go
+++ b/sdktests/common_tests_events_contexts.go
@@ -192,7 +192,7 @@ func (c CommonEventTests) EventContexts(t *ldtest.T) {
 
 			initialContext := contexts.NextUniqueContext()
 			client := NewSDKClient(t, c.baseSDKConfigurationPlus(
-				WithClientSideConfig(servicedef.SDKConfigClientSideParams{InitialContext: initialContext}),
+				WithClientSideInitialContext(initialContext),
 				WithEventsConfig(p.eventsConfig),
 				dataSource,
 				events)...)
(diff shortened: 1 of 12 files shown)


sdktests/common_tests_events_contexts.go
```diff
@@ -252,7 +252,7 @@ func (c CommonEventTests) EventContexts(t *ldtest.T) {
 					eventMatchers := []m.Matcher{debugEventMatcher, IsSummaryEvent()}
 					if c.isClientSide {
 						eventMatchers = append(eventMatchers, IsIdentifyEvent())
-					} else if !c.isPHP {
+					} else {
```
eli-darkly, line 255: The condition here was redundant because we already verified it's not PHP in the enclosing block.

sdktests/common_tests_events_contexts.go
```diff
@@ -265,7 +265,11 @@ func (c CommonEventTests) EventContexts(t *ldtest.T) {
 							_ = basicEvaluateFlagWithOldUser(t, client, debuggedFlagKey, user, defaultValue)
 							client.FlushEvents(t)
 							payload := events.ExpectAnalyticsEvents(t, defaultEventTimeout)
-							m.In(t).Assert(payload, m.ItemsInAnyOrder(debugEventMatcher, IsSummaryEvent()))
```
eli-darkly, line 268: This expectation was wrong for client-side SDKs because we have explicitly sent an identify event in this payload.

cwaldren-ld (approved):

PR #122, opened 2023-01-30
Add environment filtering tests

cwaldren-ld:
This PR introduces optional tests for filtered environment functionality. 

The framework passes in a new configuration option on the streaming/polling configs, named `Filter`. 

An SDK should setup its streaming/polling data source using this filter, and then append it to requests as a URL query parameter (example: `?filter=microservice-1`).

The tests are only run for server-side SDKs with the `filtering` capability.

- [x] Go Server SDK passes tests (locally)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 .vscode/
 bin/
 dist/
-
+.idea
 sdk-test-harness
--- a/sdktests/custom_matchers_general.go
+++ b/sdktests/custom_matchers_general.go
@@ -22,10 +22,10 @@ import (
 // The functions in this file are for convenient use of the matchers API with complex
 // types. For more information, see matchers.Transform.
 
-// QueryParameters returns a MatcherTransform which parses a string representing a URL's
+// UniqueQueryParameters returns a MatcherTransform which parses a string representing a URL's
 // RawQuery field into a map from parameter key to parameter value. If there are multiple values
-// for a key, the first is used.
-func QueryParameters() m.MatcherTransform {
+// for a key, an error is returned.
+func UniqueQueryParameters() m.MatcherTransform {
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -1,5 +1,5 @@
 .vscode/
 bin/
 dist/
-
+.idea
```
kinyoklion, line 4: :tada:

kinyoklion:
PR description says `filters` capability, the code says `filtering`. Do we want it to be `filters` or `filtering`?

cwaldren-ld:
Fixed! (`filtering`)

sdktests/custom_matchers_general.go
```diff
@@ -21,6 +22,23 @@ import (
 // The functions in this file are for convenient use of the matchers API with complex
 // types. For more information, see matchers.Transform.
 
+// QueryParameters returns a MatcherTransform which parses a string representing a URL's
+// RawQuery field into a map from parameter key to parameter value. If there are multiple values
+// for a key, the first is used.
```
kinyoklion, line 27: Maybe multiple values could be an error? It feels problematic for testing if you make a  URL with something like `filter=potato&filter=bacon` that if the desired filter was `potato` that it would pass?

cwaldren-ld: I missed that when I refactored to use a MatcherTransform here, will fix.

kinyoklion (approved):

cwaldren-ld:
Note: 

These tests are missing an area of coverage, which is when an SDK reconnects to the stream. The filter parameter should still be present in reconnects. 

There are a lot of tests for reoconnection, and they setup the SDK in a bunch of different ways - I think it'd be rather messy to add in assertions in all those places.

I'll look for a way to accomplish "every request made by this SDK should have parameter X" more elegantly.

PR #129, opened 2023-03-22
Add support for Roku SDK

keelerm84:
The Roku platform lacks sufficient support for a streaming HTTPS connection. As a result, we have separate dedicated endpoints for serving the Roku SDK.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,6 @@ require (
 	github.com/mattn/go-colorable v0.1.9 // indirect
 	github.com/mattn/go-isatty v0.0.14 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	golang.org/x/sys v0.0.0-20211019181941-9d821ace8654 // indirect
+	golang.org/x/sys v0.1.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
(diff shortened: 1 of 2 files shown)


mockld/crypto/config.go
```diff
@@ -0,0 +1,6 @@
+package crypto
```
keelerm84, line 1: Everything in crypto was copied from streamer (changing only the imports and removing some feature flagging stuff we don't need).

mockld/streaming_service.go
```diff
@@ -80,6 +84,16 @@ func NewStreamingService(
 	switch sdkKind {
 	case ServerSideSDK:
 		router.HandleFunc(StreamingPathServerSide, streamHandler).Methods("GET")
+	case RokuSDK:
+		config := crypto.CryptoConfig{
+			AuthKey:   "YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXowMTIzNDU=",
+			CipherKey: "YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXowMTIzNDU=",
```
keelerm84, line 89: Wasn't sure what we wanted to do here. It just needs to be 2 32-character strings, and then base64 encoded. These are a-z0-5. :shrug:

PR #136, opened 2023-06-08
feat: Add polling test with large payload size

keelerm84:
We recently discovered a bug where one of our SDKs wasn't processing the
full payload message. To help combat that, we are introducing a new test
which contains a large payload.

### DIFF
--- a/sdktests/common_tests_poll_request.go
+++ b/sdktests/common_tests_poll_request.go
@@ -37,11 +37,11 @@ func (c CommonPollingTests) RequestMethodAndHeaders(t *ldtest.T, credential stri
 }
 
 func (c CommonPollingTests) LargePayloads(t *ldtest.T) {
-	flags := make([]ldmodel.FeatureFlag, 100)
-	for i := 0; i < 100; i++ {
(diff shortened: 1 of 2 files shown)


sdktests/common_tests_poll_request.go
```diff
@@ -33,6 +36,53 @@ func (c CommonPollingTests) RequestMethodAndHeaders(t *ldtest.T, credential stri
 	})
 }
 
+func (c CommonPollingTests) LargePayloads(t *ldtest.T) {
+	flags := make([]ldmodel.FeatureFlag, 100)
```
cwaldren-ld, line 40: I'd be curious to know the size of this in bytes. Like - is this 1 kb? 10kb? Should we be using 1000 flags?

cwaldren-ld (approved):

sdktests/common_tests_poll_request.go
keelerm84: 100 flags was ~59k. Jumped up to 1,000 just to be safe. Now we are around 585k.

PR #149, opened 2023-08-16
feat: Add initial migration related tests

keelerm84:
These initial tests cover the following areas:

1. Ensure migration variation method returns the correct stage.
2. Ensure migrations execute calls from old and new origins for the appropriate stage.
3. Ensure latency and error metrics are present when enabled.

### DIFF
--- a/sdktests/server_side_migrations.go
+++ b/sdktests/server_side_migrations.go
@@ -29,7 +29,7 @@ func doServerSideMigrationTests(t *ldtest.T) {
 }
 
 func runMigrationVariationTests(t *ldtest.T) {
-	stages := []ldmigration.MigrationStage{ldmigration.Off, ldmigration.DualWrite, ldmigration.Shadow, ldmigration.Live, ldmigration.RampDown, ldmigration.Complete}
+	stages := []ldmigration.Stage{ldmigration.Off, ldmigration.DualWrite, ldmigration.Shadow, ldmigration.Live, ldmigration.RampDown, ldmigration.Complete}
 
 	for _, stage := range stages {
 		client, events := createClient(t, int(stage))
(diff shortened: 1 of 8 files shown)


sdktests/server_side_migrations.go
```diff
@@ -0,0 +1,370 @@
...
+func doServerSideMigrationTests(t *ldtest.T) {
+	t.RequireCapability(servicedef.CapabilityMigrations)
+
+	t.Run("migration variation", runMigrationVariationTests)
+	t.Run("use correct origins", runUseCorrectOriginsTests)
+	t.Run("track latency correctly", runTrackLatencyTests)
+	t.Run("track errors correctly", runTrackErrorsTests)
+}
+
+func runMigrationVariationTests(t *ldtest.T) {
```
keelerm84, line 29: Just making sure a basic migration flag will return the right stage.

sdktests/server_side_migrations.go
```diff
@@ -0,0 +1,370 @@
...
+			for idx, callHistory := range service.GetCallHistory() {
+				assert.Equal(t, testParam.ExpectedRequests[idx], callHistory.GetOrigin())
+			}
+
+			assert.Equal(t, testParam.ExpectedResult, response.Result)
+		})
+	}
+}
+
+func runTrackLatencyTests(t *ldtest.T) {
```
keelerm84, line 132: Basic latency tests. Just making sure something was provided.

sdktests/server_side_migrations.go
```diff
@@ -0,0 +1,370 @@
...
+				IsValidMigrationOpEventWithConditions(
+					context,
+					opEventMatchers...,
+				),
+			))
+		})
+	}
+}
+
+func runTrackErrorsTests(t *ldtest.T) {
```
keelerm84, line 229: Because authoritative write failures should halt execution, testing for error tracking has many more cases than latency tracking.

keelerm84 (commented):
There are still more tests to write, particularly around event sampling. But this should mostly cover the basic functionality.

Documentation also needs updated, but since I'm just merging this into a feature branch for now, I feel comfortable doing that in a follow-up task.

sdktests/server_side_migrations.go
```diff
@@ -50,7 +50,7 @@ func runMigrationVariationTests(t *ldtest.T) {
 			IsSummaryEvent(),
 		))
 
-		assert.Equal(t, stage, response.Result)
+		assert.Equal(t, string(stage), response.Result)
```
cwaldren-ld, line 53: There's some assertion that will do this for you. I think maybe.. `assert.EqualValues`?

cwaldren-ld (approved):

PR #150, opened 2023-08-17
fix: relaxing context type test, now contains application/json

tanderson-ld:
Java SDK was failing on this test, just relaxing the test a smidge.


cwaldren-ld (approved):

cwaldren-ld:
Is there some spec update/clarification related to this that needs to happen? Or was this just wrong?

tanderson-ld:
> 

I think it was just overly strict.  Other content type tags may exist simultaneously I believe.  Not an expert on the HTTP specs though.

PR #152, opened 2023-08-24
Updated tags tests to account for fallback when id is invalid

tanderson-ld:
Updating ApplicationInfo tag tests to allow spaces and to test fallback behavior.

### DIFF
--- a/sdktests/constants.go
+++ b/sdktests/constants.go
@@ -1,7 +1,10 @@
 package sdktests
 
 const (
-	allAllowedTagChars = "._-abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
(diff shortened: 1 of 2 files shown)


sdktests/constants.go
```diff
@@ -1,7 +1,10 @@
 package sdktests
 
 const (
-	allAllowedTagChars = "._-abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	// Application tags send to the cloud are not allowed to have spaces.  However, the SDKs
```
keelerm84, line 4:
```suggestion
	// Application tags sent to the cloud are not allowed to have spaces.  However, the SDKs
```

keelerm84 (approved):

PR #153, opened 2023-08-28
feat: Flag and custom sampling ratios and flag summary tests

keelerm84:


mockld/sdk_data.go
```diff
@@ -166,27 +166,46 @@ func normalizeSegment(key string, data json.RawMessage) (json.RawMessage, error)
...
+		metrics:         make(map[string]json.RawMessage),
 	}
 }
 
 func (b *ServerSDKDataBuilder) Build() ServerSDKData {
 	flags := make(map[string]json.RawMessage)
 	segments := make(map[string]json.RawMessage)
+	configOverrides := make(map[string]json.RawMessage)
+	metrics := make(map[string]json.RawMessage)
+
```
cwaldren-ld, line 189: I wonder.. can we replace all of these for loops with [`map.Clone`](https://pkg.go.dev/golang.org/x/exp/maps#Clone) now?

cwaldren-ld (approved):

mockld/sdk_data.go
keelerm84: Yep, that seems to work just fine.

PR #155, opened 2023-08-30
fix: Relaxing content-type for server events.

kinyoklion:
We already did this for mobile, but it is needed for java server as well.


sdktests/server_side_events_all.go
```diff
@@ -33,7 +33,7 @@ func doServerSideEventRequestTests(t *ldtest.T) {
 	eventTests.RequestMethodAndHeaders(t, sdkKey, m.AllOf(
 		Header("X-LaunchDarkly-Event-Schema").Should(m.Equal(currentEventSchema)),
 		Header("X-LaunchDarkly-Payload-Id").Should(m.Not(m.Equal(""))),
-		Header("Content-Type").Should(m.Equal("application/json")),
+		Header("Content-Type").Should(m.StringContains("application/json")),
```
kinyoklion, line 36: Java will include a `charset`. This is not required for `application/json` (because it is UTF-8 by its own standard), but it isn't explicitly forbidden.

The java header will be.
`application/json; charset=utf-8`

tanderson-ld (approved):

PR #156, opened 2023-08-30
feat: Add tests verifying payload behavior

keelerm84:


keelerm84:
This should be reviewed one commit at a time. The first is the dependency bump and path changes. You can basically ignore that one. The second one is the changes required as a result of the version bump. Then the final commit is the actual new payload tests.

cwaldren-ld (approved):

PR #169, opened 2023-10-18
feat: Add PHP support for migration related tests

keelerm84:


sdktests/server_side_migrations.go
```diff
@@ -49,6 +49,10 @@ func withExecutionOrders(test func(*ldtest.T, ldmigration.ExecutionOrder)) func(
 
 	return func(t *ldtest.T) {
 		for _, order := range executionOrders {
+			if t.Capabilities().Has(servicedef.CapabilityPHP) && order == ldmigration.Concurrent {
+				continue
```
kinyoklion, line 53: Should we have a `t.Skip()`, maybe with a reason.

kinyoklion (approved):

sdktests/server_side_migrations.go
keelerm84: Adding a `t.Skip()` here would abort the entire test instead of only skipping the concurrent branch. So I don't think I can actually do that.

PR #173, opened 2023-12-18
feat: Add capability and test for context comparison

keelerm84:


docs/service_spec.md
```diff
@@ -315,6 +319,27 @@ The `contextConvert` property in the request body will be a JSON object with the
...
+The test harness will only send this command if the test service has the `"context-comparison"` capability.
+
+The `contextComparison` property in the request body will be a JSON object with these properties:
+
+* `context1` (object): One of two contexts which should be constructed and compared.
+  * `single` (object, optional): If present, this is a JSON object with properties for a single-kind context. The test service should pass these values to the corresponding builder methods if they are present.
+    * `kind` (string, optional): Even though a context always has a kind, this is optional because the builder should use `"user"` as a default.
+    * `key` (string, required)
+    * `properties` (array, optional): If present, this contains an array of context property definitions. If the SDK has a builder for a context, these should be applied in the order sent. Each property definition contains a name and value field.
+    * `privateAttributes` (array, optional): if present, this contains an array of private attribute definitions. Each attribute has a `literal` field to designate how the `value` property should be interpreted.
```
cwaldren-ld, line 335: Perhaps we could add something like: "If `literal` is true, the `value` is an attribute name, rather than an attribute reference. For example, `{"value" : "/foo/bar", "literal" : "true"}` means the attribute named `/foo/bar` rather than a nested object named `bar` under the `foo` object.

docs/service_spec.md
```diff
@@ -315,6 +319,27 @@ The `contextConvert` property in the request body will be a JSON object with the
...
+#### Comparison contexts for equality
+
+If `command` is `"contextComparison"`, the test service should construct two separate contexts using the definitions provided in the payload. These contexts should then be compared for equality.
+
+The test harness will only send this command if the test service has the `"context-comparison"` capability.
+
+The `contextComparison` property in the request body will be a JSON object with these properties:
+
+* `context1` (object): One of two contexts which should be constructed and compared.
+  * `single` (object, optional): If present, this is a JSON object with properties for a single-kind context. The test service should pass these values to the corresponding builder methods if they are present.
```
cwaldren-ld, line 331:
```suggestion
* `context1` (object): One of two contexts which should be constructed and compared. Contains a single field, `single` or `multi`.
  * `single` (object, optional): If present, this is a JSON object with properties for a single-kind context. The test service should pass these values to the corresponding builder methods if they are present.
```

servicedef/command_params.go
```diff
@@ -118,6 +120,37 @@ type ContextConvertParams struct {
...
+
+type ContextComparisonParams struct {
+	Single *ContextComparisonSingleParams  `json:"single,omitempty"`
+	Multi  []ContextComparisonSingleParams `json:"multi,omitempty"`
+}
+
+type ContextComparisonSingleParams struct {
+	Kind              string               `json:"kind"`
+	Key               string               `json:"key"`
+	Properties        []PropertyDefinition `json:"properties,omitempty"`
```
cwaldren-ld, line 136:
```suggestion
	Attributes        []AttributeDefinition `json:"attributes,omitempty"`
```

servicedef/command_params.go
```diff
@@ -118,6 +120,37 @@ type ContextConvertParams struct {
...
+	PrivateAttributes []PrivateAttribute   `json:"privateAttributes:omitempty"`
+}
+
+type PropertyDefinition struct {
```
cwaldren-ld, line 140:
```suggestion
type AttributeDefinition struct {
```

cwaldren-ld (approved):

PR #178, opened 2024-01-17
feat: Add support for anonymous attribute redaction in feature events

keelerm84:


mockld/roku.go
```diff
@@ -28,7 +28,7 @@ func (srv *RokuServer) Wrap(h http.Handler) http.Handler {
 			return
 		}
 
-		userJSONBase64 := base64.StdEncoding.EncodeToString(userJSON)
+		userJSONBase64 := base64.URLEncoding.EncodeToString(userJSON)
```
cwaldren-ld, line 31: is this an unrelated bugfix?

sdktests/client_side_events_eval.go
```diff
@@ -127,6 +129,133 @@ func doClientSideFeatureEventTests(t *ldtest.T) {
...
+			t.DebugLogger().Println(2)
+
+			for _, valueType := range getValueTypesToTest(t) {
+				t.Run(testDescFromType(valueType), func(t *ldtest.T) {
+					t.DebugLogger().Println(3)
+					userContext := userContextFactory.NextUniqueContext()
+					orgContext := orgContextFactory.NextUniqueContext()
+
+					multiContext := ldcontext.NewMultiBuilder().Add(userContext).Add(orgContext).Build()
+					t.DebugLogger().Println(4)
```
cwaldren-ld, line 206: Should the debug logs remain?

mockld/roku.go
keelerm84: That other PR landed the fix on v2 so we can get the benefit immediately. I had to wait for it to land, pull those changes into the feat/inline-contexts branch, and then update this PR accordingly.

tl;dr -- removed!

sdktests/client_side_events_eval.go
keelerm84: Nope and removed.

cwaldren-ld (approved):

PR #179, opened 2024-01-17
fix: Use correct payload type for server/client in common tests.

kinyoklion:
Some tests common to client/server use a server payload instead of a client payload.


sdktests/common_tests_events_custom.go
```diff
@@ -38,7 +37,7 @@ func (c CommonEventTests) CustomEvents(t *ldtest.T) {
 			t.Run(contexts.Description(), func(t *ldtest.T) {
 				context := contexts.NextUniqueContext()
 
-				dataSource := NewSDKDataSource(t, mockld.EmptyServerSDKData())
+				dataSource := NewSDKDataSource(t, nil)
```
kinyoklion, line 40: If you do not specify the data, then it will try to make data for the correct SDK type based on the context.

tanderson-ld (approved):

tanderson-ld (approved):

PR #181, opened 2024-01-18
feat: Support PHP v2 and v4 schema payloads

keelerm84:


sdktests/php_events_all.go
```diff
@@ -34,7 +34,14 @@ func doPHPEventRequestTests(t *ldtest.T) {
 		}))
 
 	eventTests.RequestMethodAndHeaders(t, sdkKey,
-		Header("X-LaunchDarkly-Event-Schema").Should(m.Equal(phpEventSchema)))
```
cwaldren-ld, line 37: In this case, should `phpEventSchema` be renamed something like `php(Deprecated|Legacy|Whatever)EventSchema` to indicate that it's no longer the latest and greatest?

sdktests/php_events_all.go
```diff
@@ -34,7 +34,14 @@ func doPHPEventRequestTests(t *ldtest.T) {
...
+		m.AnyOf(
+			Header("X-LaunchDarkly-Event-Schema").Should(m.Equal(phpEventSchema)),
+			m.AllOf(
+				Header("X-LaunchDarkly-Event-Schema").Should(m.Equal(currentEventSchema)),
+				Header("X-LaunchDarkly-Summarized").Should(m.Not(m.BeNil())),
```
cwaldren-ld, line 41: Should be `X-LaunchDarkly-Unsummarized ?

sdktests/php_events_all.go
keelerm84: Done!

sdktests/php_events_all.go
keelerm84: Done

cwaldren-ld (approved):

PR #182, opened 2024-01-19
feat: Add php support for anonymous attribute redaction in feature events

keelerm84:


sdktests/common_tests_events_contexts.go
```diff
@@ -51,7 +51,6 @@ func makeEventContextTestParams() []eventContextTestParams {
 					b.Kind("org")
 					b.Name("a")
 					b.SetString("b", "c")
-					b.Anonymous(true)
```
cwaldren-ld, line 54: Why this change?

sdktests/php_events_eval.go
```diff
@@ -150,6 +150,123 @@ func doPHPFeatureEventTests(t *ldtest.T) {
...
+
+			for _, valueType := range getValueTypesToTest(t) {
+				t.Run(testDescFromType(valueType), func(t *ldtest.T) {
+					flag := flagFactories[flagSelectors{
+						tracked:   true,
+						withDebug: false,
+						malformed: false,
+					}].ReuseFlagForValueType(valueType)
+					var expectedValue ldvalue.Value
+					// var expectedVariation o.Maybe[int]
```
cwaldren-ld, line 170: commented code

sdktests/common_tests_events_contexts.go
keelerm84: These tests are used for PHP and non-PHP server side SDKs.

If the SDK is PHP, it takes a branch that tries to validate the context provided in that event. You can change what that expectation looks like but it also affects the index event.

Non-PHP doesn't have this same branch. By removing the anonymous flag, the feature event context won't be different from the index one, and I don't have to change a ton of stuff.

Also, I have a test verifying the anonymous behavior of PHP with feature events, so no coverage is lost.

sdktests/php_events_eval.go
keelerm84: Done!

cwaldren-ld (approved):

PR #185, opened 2024-01-30
feat: Add test verifying gzip behavior for polling

keelerm84:


tanderson-ld (approved):

mockld/polling_service.go
```diff
@@ -117,8 +122,22 @@ func (p *PollingService) pollingHandler(getDataFn func(*PollingService, *http.Re
...
 			w.Header().Add("Etag", etag)
 		}
-		w.WriteHeader(http.StatusOK)
-		_, _ = w.Write(data)
+
+		if p.enableGzipCompression && strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
+			w.Header().Add("Content-Encoding", "gzip")
+			w.WriteHeader(http.StatusOK)
+			gzipWriter := gzip.NewWriter(w)
+			_, _ = gzipWriter.Write(data)
```
cwaldren-ld, line 130: I think the `Write` error(s) should be inspected/logged here. If the write or flush fails it might be confusing why.

cwaldren-ld (approved):

mockld/polling_service.go
```diff
@@ -38,23 +40,26 @@ const (
...
+	enableGzipCompression bool
+	debugLogger           framework.Logger
+	lock                  sync.RWMutex
 }
 
 func NewPollingService(
 	initialData SDKData,
 	sdkKind SDKKind,
+	enableGzipCompression bool,
 	debugLogger framework.Logger,
```
cwaldren-ld, line 56: I think it could be a good time to refactor this to be more go-ish, with optional arguments via adding a `WithGzipCompression(bool)` method to `*PollingService`. So:
```golang
p := NewPollingService(x, y, z).WithGzipCompression(true)
```

cwaldren-ld (approved):

PR #188, opened 2024-02-20
ci: add release-please config and workflows [v2]

cwaldren-ld:
This enables `release-please` on the v2 branch. I've taken a slightly different approch to the config from our template, just to see if it's nicer.

Instead of duplicating the `ci + publish` steps in the `manual-publish` and `release-please` jobs, I stuck it in a reusable workflow that both can call.


.github/actions/publish/action.yml
```diff
@@ -0,0 +1,30 @@
...
+  tag:
+    description: 'Tag to upload artifacts to.'
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Git clean
+      shell: bash
+      run: git clean -f
```
tanderson-ld, line 16: What is the reason for this clean?  Feels like cleans in actions could lead to unexpected results when combined into a workflow.  Or make them difficult to move as more workflows/actions get added over time.

I use `git clean -xdf` locally sometimes, consider the x and d arguments?

tanderson-ld (approved):

.github/actions/publish/action.yml
cwaldren-ld: The clean is because `goreleaser`  will fail if git is in a dirty state. Since the publish step is run after unit tests, there's usually a coverage file or similar created. 

I think I agree with what you're getting at though - this is a brute force way of doing it; the publish action is acting defensively. It might be better to have the unit test step ensure git is clean instead.

PR #189, opened 2024-02-21
chore: fix secret passing in release workflows

cwaldren-ld:
Didn't realize that reusable workflows need a special syntax for passing secrets like `GITHUB_TOKEN`, different than `inputs`. 

The existing config results in an invalid workflow file (although it works fine in `act`.)

### DIFF
--- a/.github/workflows/manual-release.yml
+++ b/.github/workflows/manual-release.yml
@@ -11,5 +11,6 @@ jobs:
   manual-release:
     uses: ./.github/workflows/release.yml
     with:
-      token: ${{ secrets.GITHUB_TOKEN }}
       tag: ${{ inputs.tag }}
+    secrets:
+      token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/manual-release.yml
+++ b/.github/workflows/manual-release.yml
@@ -12,5 +12,3 @@ jobs:
     uses: ./.github/workflows/release.yml
     with:
       tag: ${{ inputs.tag }}
-    secrets:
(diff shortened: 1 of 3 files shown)


.github/workflows/manual-release.yml
```diff
@@ -11,5 +11,6 @@ jobs:
   manual-release:
     uses: ./.github/workflows/release.yml
     with:
-      token: ${{ secrets.GITHUB_TOKEN }}
       tag: ${{ inputs.tag }}
+    secrets:
+      token: ${{ secrets.GITHUB_TOKEN }}
```
kinyoklion, line 16: Because it is a workflow, instead of a composite, you shouldn't need to pass. The workflow can have its own permissions.

kinyoklion: https://github.com/launchdarkly/dotnet-logging-adapter-nlog/blob/3c92a9e0ad1c64a2a327ed2bffab3ca25e3906f8/.github/workflows/release-please.yml#L27

kinyoklion:
```
When a reusable workflow is triggered by a caller workflow, the github context is always associated with the caller workflow. The called workflow is automatically granted access to github.token and secrets.GITHUB_TOKEN. For more information about the github context, see "[Contexts](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)."
```

cwaldren-ld: Yea, I see that `secrets` is automatically available. This does work though, but I'll refactor to remove it.

kinyoklion (approved):

PR #191, opened 2024-02-21
fix(deps): build with Go 1.22

cwaldren-ld:
This PR builds sdk-test-harness with Go 1.22, up from Go 1.18. With this change, we can pull in a new `golanglintci` version. 

This surfaced a bunch of new lints, which I've fixed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,3 +1,5 @@
+name: Publish Release
+
 on:
   workflow_call:
     inputs:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,7 +5,6 @@ run:
 linters:
   enable:
     - bodyclose
-    - depguard
     - dupl
     - errcheck
     - gochecknoglobals
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,7 +5,6 @@ run:
 linters:
   enable:
     - bodyclose
-    - deadcode
     - depguard
     - dupl
     - errcheck
@@ -27,13 +26,11 @@ linters:
     - nolintlint
     - prealloc
     - staticcheck
-    - structcheck
     - stylecheck
     - typecheck
     - unconvert
     - unparam
     - unused
-    - varcheck
     - whitespace
   fast: false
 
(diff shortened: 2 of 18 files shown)


.github/workflows/release.yml
```diff
@@ -1,3 +1,5 @@
+name: Publish Release
+
```
cwaldren-ld, line 2: Sneaking this in; without it it shows the full workflow file name in the UI.

.golangci.yml
```diff
@@ -5,8 +5,6 @@ run:
 linters:
   enable:
     - bodyclose
```
cwaldren-ld, line 7: deadcode and depguard are unmaintained/deprecated

.golangci.yml
```diff
@@ -24,18 +22,14 @@ linters:
     - megacheck
     - misspell
     - nakedret
```
cwaldren-ld, line 24: nolintlint works fine with go1.22

.golangci.yml
cwaldren-ld: Removed `depguard` because I don't know why we were using it (and it wasn't configured.) The purpose is to restrict the module from importing certain packages; we don't have such a policy for this tool as far as I know.

data/context_factory.go
```diff
@@ -100,11 +100,11 @@ func (f *ContextFactory) SetKeyDisambiguatorValueSameAs(f1 *ContextFactory) {
 // Each will have an appropriate Description, so the logic for running a test against each one can look
```
cwaldren-ld, line 100: (gofmt)

sdktests/common_tests_events_contexts.go
```diff
@@ -227,7 +227,7 @@ func (c CommonEventTests) EventContexts(t *ldtest.T) {
 
 					if user := representContextAsOldUser(t, context); user != nil {
 						t.Run("with old user", func(t *ldtest.T) {
-							_ = basicEvaluateFlagWithOldUser(t, client, flagKey, user, defaultValue)
```
kinyoklion, line 230: Not related to this change, but this is some non-intuitive data flow.

kinyoklion (approved):

sdktests/common_tests_events_contexts.go
cwaldren-ld: Agreed.

PR #193, opened 2024-02-21
ci: don't specify release type on release-please action

cwaldren-ld:
The [release PR ](https://github.com/launchdarkly/sdk-test-harness/pull/192) didn't update [`main.go`](https://github.com/launchdarkly/sdk-test-harness/blob/v2/main.go#L20) as expected - I'm not seeing `fetching main.go from branch v2` or similar in the logs.

It's possible this is because the release-type is `simple`, when I meant it to be `go`. 

This may or may not solve the problem, but in any case it's a necessary change.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -3,7 +3,7 @@
     "." : {
       "release-type" : "go",
       "bump-minor-pre-major" : true,
-      "boostrap-sha" : "12774d23187e79d802083368132ea95b64a45cca",
+      "boostrap-sha" : "b244281d769f355998ec7e7c9da7f15344d8c92d",
       "versioning" : "default",
       "include-component-in-tag" : false,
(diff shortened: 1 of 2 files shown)


release-please-config.json
```diff
@@ -3,7 +3,7 @@
     "." : {
       "release-type" : "go",
       "bump-minor-pre-major" : true,
-      "boostrap-sha" : "12774d23187e79d802083368132ea95b64a45cca",
+      "boostrap-sha" : "b244281d769f355998ec7e7c9da7f15344d8c92d",
       "versioning" : "default",
```
cwaldren-ld, line 7: Setting this to the oldest conventionally-commited commit now that we have a few.

louis-launchdarkly (approved):

PR #200, opened 2024-03-14
feat: Add hooks contract tests.

kinyoklion:
This PR adds support for basic contract testing of hooks.

It does not include support for dynamic registration of hooks.
It does not yet test hook execution order.
It does not test hooks throwing errors.

### DIFF
--- a/docs/service_spec.md
+++ b/docs/service_spec.md
@@ -141,6 +141,9 @@ A test hook must:
         * `method` (string, required): The name of the evaluation emthod that was called.
       * `evaluationHookData` (object, optional): The EvaluationHookData passed to the stage during execution.
       * `evaluationDetail` (object, optional): The details of the evaluation if executing an `afterEvaluation` stage.
+        * `value` (any): The JSON value of the result.
+        * `variationIndex` (int or null): The variation index of the result.
+        * `reason` (object): The evaluation reason of the result.
       * `stage` (string, optional): If executing a stage, for example `beforeEvaluation`, this should be the stage.
   - Return data from the stages as specified via the `data` configuration. For instance the return value from the `beforeEvaluation` hook should be `data['beforeEvaluation']` merged with the input data for the stage.
 
--- a/docs/service_spec.md
+++ b/docs/service_spec.md
@@ -134,12 +134,12 @@ A test hook must:
   - Implement the SDK hook interface.
   - Whenever an evaluation stage is called post information about that call to the `callbackUrl` of the hook.
     - The payload is an object with the following properties:
-      * `evaluationHookContext` (object, optional): If an evaluation stage was executed, then this should be the associated context.
+      * `evaluationSeriesContext` (object, optional): If an evaluation stage was executed, then this should be the associated context.
         * `flagKey` (string, required): The key of the flag being evaluated.
         * `context` (object, required): The evaluation context associated with the evaluation.
         * `defaultValue` (any): The default value for the evaluation.
         * `method` (string, required): The name of the evaluation emthod that was called.
-      * `evaluationHookData` (object, optional): The EvaluationHookData passed to the stage during execution.
+      * `evaluationSeriesData` (object, optional): The EvaluationSeriesData passed to the stage during execution.
       * `evaluationDetail` (object, optional): The details of the evaluation if executing an `afterEvaluation` stage.
       * `stage` (string, optional): If executing a stage, for example `beforeEvaluation`, this should be the stage.
   - Return data from the stages as specified via the `data` configuration. For instance the return value from the `beforeEvaluation` hook should be `data['beforeEvaluation']` merged with the input data for the stage.
--- a/docs/service_spec.md
+++ b/docs/service_spec.md
@@ -124,6 +124,26 @@ and will send a `?filter=name` query parameter along with streaming/polling requ
 For tests that involve filtering, the test harness will set the `filter` property of the `streaming` or `polling` configuration
 object. The property will either be omitted if no filter is requested, or a non-empty string if requested.
 
+### Capability `"evaluation-hooks"`
+
+This means that the SDK has support for hooks and has the ability to register evaluation hooks.
+
+For a test service to support hooks testing it must support a `test-hook`. The configuration will specify registering one or more `test-hooks`.
+
+A test hook must:
+  - Implement the SDK hook interface.
+  - Whenever an evaluation stage is called post information about that call to the `callbackUrl` of the hook.
+    - The payload is an object with the following properties:
+      * `evaluationHookContext` (object, optional): If an evaluation stage was executed, then this should be the associated context.
+        * `flagKey` (string, required): The key of the flag being evaluated.
+        * `context` (object, required): The evaluation context associated with the evaluation.
+        * `defaultValue` (any): The default value for the evaluation.
+        * `method` (string, required): The name of the evaluation emthod that was called.
+      * `evaluationHookData` (object, optional): The EvaluationHookData passed to the stage during execution.
+      * `evaluationDetail` (object, optional): The details of the evaluation if executing an `afterEvaluation` stage.
+      * `stage` (string, optional): If executing a stage, for example `beforeEvaluation`, this should be the stage.
+  - Return data from the stages as specified via the `data` configuration. For instance the return value from the `beforeEvaluation` hook should be `data['beforeEvaluation']` merged with the input data for the stage.
+
 ### Stop test service: `DELETE /`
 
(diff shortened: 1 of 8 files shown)


sdktests/server_side_hooks.go
```diff
@@ -0,0 +1,307 @@
...
+			flagKey:      "number-flag",
+			defaultValue: ldvalue.Int(0xDEADBEEF),
+			valueType:    servicedef.ValueTypeInt,
+			detail:       detail,
+		},
+		{
+			name:         "for json variation",
+			flagKey:      "json-flag",
+			defaultValue: ldvalue.ObjectBuild().Build(),
+			valueType:    servicedef.ValueTypeInt,
```
cwaldren-ld, line 87: Intend for this to be ValueTypeInt?

sdktests/server_side_hooks.go
```diff
@@ -0,0 +1,307 @@
...
+	for _, testParam := range testParams {
+		t.Run(testParam.name, func(t *ldtest.T) {
+			client.EvaluateFlag(t, servicedef.EvaluateFlagParams{
+				FlagKey:      testParam.flagKey,
+				Context:      o.Some(ldcontext.New("user-key")),
+				ValueType:    testParam.valueType,
+				DefaultValue: testParam.defaultValue,
+			})
+
+			hooks.ExpectCall(t, hookName, 1*time.Second, func(payload servicedef.HookExecutionPayload) bool {
```
cwaldren-ld, line 109: This 1 second should be pulled out into a constant (or perhaps use one of the existing timeout constants?)

sdktests/server_side_hooks.go
kinyoklion: I did not. It also indicates I need to add a value check.

kinyoklion: Actually. I need to figure out why it passed, because I do have one.

kinyoklion: Ah, yeah, it doesn't have an independent evaluation. That is probably fine though, this more cares that they match than testing the evaluation itself.

sdktests/server_side_hooks.go
```diff
@@ -0,0 +1,307 @@
...
+
+	flagKey := "migration-flag"
+	params := servicedef.MigrationVariationParams{
+		Key:          flagKey,
+		Context:      ldcontext.New("user-key"),
+		DefaultStage: ldmigration.Off,
+	}
+	result := client.MigrationVariation(t, params)
+
+	hooks.ExpectCall(t, hookName, 1*time.Second, func(payload servicedef.HookExecutionPayload) bool {
```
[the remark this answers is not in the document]
kinyoklion: I am not sure. I think, for the existing tests, that would be a little simpler. 

If you wanted to do something with multiple calls, until some condition was met, then it wouldn't work out as well. But maybe that isn't a good idea.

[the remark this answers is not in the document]
kinyoklion: Maybe more reasonable is you want a condition other than the stage.

mockld/hook_callback_service.go
```diff
@@ -0,0 +1,63 @@
...
+		logger.Printf("Received from hook: %s", string(bytes))
+		if err != nil {
+			logger.Printf("Could not read body from hook.")
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		var response servicedef.HookExecutionPayload
+		err = json.Unmarshal(bytes, &response)
+		if err != nil {
+			logger.Printf("Could not unmarshall hook payload.")
```
cwaldren-ld, line 47:
```suggestion
			logger.Printf("Could not unmarshal hook payload.")
```

cwaldren-ld:
Curious why the change to the non-blocking push? I'm guessing it was forcing the SDK to be synchronized with whatever test code expectations had (or had not) been written?

kinyoklion:
> Curious why the change to the non-blocking push? I'm guessing it was forcing the SDK to be synchronized with whatever test code expectations had (or had not) been written?

The node SDK doesn't have a way to do a synchronous http request. So it fires and forgets the request and the tests passed. Because the evaluation continues.
Ruby can send it synchronously, but that blocks the evaluation (waiting for the test to accept the payload), so an evaluation call never completes.

cwaldren-ld (approved):

docs/service_spec.md
```diff
@@ -124,6 +124,26 @@ and will send a `?filter=name` query parameter along with streaming/polling requ
 For tests that involve filtering, the test harness will set the `filter` property of the `streaming` or `polling` configuration
 object. The property will either be omitted if no filter is requested, or a non-empty string if requested.
 
+### Capability `"evaluation-hooks"`
```
cwaldren-ld, line 127:
```suggestion
#### Capability `"evaluation-hooks"`
```

docs/service_spec.md
```diff
@@ -124,6 +124,26 @@ and will send a `?filter=name` query parameter along with streaming/polling requ
...
+
+This means that the SDK has support for hooks and has the ability to register evaluation hooks.
+
+For a test service to support hooks testing it must support a `test-hook`. The configuration will specify registering one or more `test-hooks`.
+
+A test hook must:
```
cwaldren-ld, line 133: This section should potentially go in the "callbacks" section at the bottom of the doc

docs/service_spec.md
```diff
@@ -124,6 +124,26 @@ and will send a `?filter=name` query parameter along with streaming/polling requ
...
+    - The payload is an object with the following properties:
+      * `evaluationHookContext` (object, optional): If an evaluation stage was executed, then this should be the associated context.
+        * `flagKey` (string, required): The key of the flag being evaluated.
+        * `context` (object, required): The evaluation context associated with the evaluation.
+        * `defaultValue` (any): The default value for the evaluation.
+        * `method` (string, required): The name of the evaluation emthod that was called.
+      * `evaluationHookData` (object, optional): The EvaluationHookData passed to the stage during execution.
+      * `evaluationDetail` (object, optional): The details of the evaluation if executing an `afterEvaluation` stage.
+      * `stage` (string, optional): If executing a stage, for example `beforeEvaluation`, this should be the stage.
+  - Return data from the stages as specified via the `data` configuration. For instance the return value from the `beforeEvaluation` hook should be `data['beforeEvaluation']` merged with the input data for the stage.
```
cwaldren-ld, line 145:
```suggestion
    - The payload is an object with the following properties:
      * `evaluationSeriesContext` (object, optional): If an evaluation stage was executed, then this should be the associated context.
        * `flagKey` (string, required): The key of the flag being evaluated.
        * `context` (object, required): The evaluation context associated with the evaluation.
        * `defaultValue` (any): The default value for the evaluation.
        * `method` (string, required): The name of the evaluation emthod that was called.
      * `evaluationSeriesData` (object, optional): The EvaluationSeriesData passed to the stage during execution.
      * `evaluationDetail` (object, optional): The details of the evaluation if executing an `afterEvaluation` stage.
      * `stage` (string, optional): If executing a stage, for example `beforeEvaluation`, this should be the stage.
  - Return data from the stages as specified via the `data` configuration. For instance the return value from the `beforeEvaluation` hook should be `data['beforeEvaluation']` merged with the input data for the stage.
```

To be consistent with naming updates. Although, now that I am reading it here, perhaps Stage was the right word instead of series - the context/data applies to all elements of the series, but is seen (by the customer) in a single stage.

docs/service_spec.md
```diff
@@ -124,6 +124,26 @@ and will send a `?filter=name` query parameter along with streaming/polling requ
...
+  - Implement the SDK hook interface.
+  - Whenever an evaluation stage is called post information about that call to the `callbackUrl` of the hook.
+    - The payload is an object with the following properties:
+      * `evaluationSeriesContext` (object, optional): If an evaluation stage was executed, then this should be the associated context.
+        * `flagKey` (string, required): The key of the flag being evaluated.
+        * `context` (object, required): The evaluation context associated with the evaluation.
+        * `defaultValue` (any): The default value for the evaluation.
+        * `method` (string, required): The name of the evaluation emthod that was called.
+      * `evaluationSeriesData` (object, optional): The EvaluationSeriesData passed to the stage during execution.
+      * `evaluationDetail` (object, optional): The details of the evaluation if executing an `afterEvaluation` stage.
```
cwaldren-ld, line 143: Is `evaluationDetail` specifically an object with `value`, `variationIndex`, and `reason` properties? Let's document that.

kinyoklion: So, I am copying the level of detail from the place where it is defined for variation detail. Things are less optional, so that is fine. But it doesn't explain the content of reason. I don't think we need to either, because it would just be replicating our data model schema here.

cwaldren-ld:
TODO: change 0xDEADBEEF to something that fits into a signed int

cwaldren-ld:
Alright, .NET server is now passing all contract tests.

PR #202, opened 2024-03-28
feat: test hook error handling

cwaldren-ld:
This introduces a test that the SDK can successfully recover from a hook stage returning an error or throwing an exception. It runs under existing capability `evaluation-hooks`.

This introduces a new `errors` object on the Hooks service definition which SDKs must inspect. It contains an error message that should be thrown/returned from a specific stage.

I've only written one test here, which is that:
- For N hooks, given an error in the `beforeEvaluation` stage, ensure the `afterEvaluation` stage still runs.

Existing SDKs are expected to fail this test until implemented. I thought about having it as a new capability, but it is part of the base spec and the reminder (seeing the test fail) should be helpful.


sdktests/testapi_hooks.go
```diff
@@ -78,3 +84,28 @@ func (h *Hooks) ExpectCall(t *ldtest.T, hookName string,
 		}
 	}
 }
+
```
cwaldren-ld, line 87: This is a somewhat janky expectation method. But, we couldn't quite think of what we'd want to see if this was more generic. 

So, I think we should feel free to refactor or replace this when the time comes. Or now, if there's a better idea.

docs/service_spec.md
```diff
@@ -194,6 +194,9 @@ A `POST` request indicates that the test harness wants to start an instance of t
       * `data` (object, optional): Contains data which should return from different execution stages.
         * `beforeEvaluation` (object, optional): A map of `string` to `ldvalue` items. This should be returned from the `beforeEvaluation` stage of the test hook.
         * `afterEvaluation` (object, optional): A map of `string` to `ldvalue` items. This should be returned from the `afterEvaluation` stage of the test hook.
+      * `errors` (object, optional): Specifies that an error should be returned/exception thrown from a stage. The error message itself is not tested by the framework at this time, as it is not a specified behavior.
+        * `beforeEvaluation` (string, optional): The error/exception message that should be generated in the `beforeEvaluation` stage of the test hook.
```
kinyoklion, line 198: Should the error be thrown after the callback post? Or should it not execute the callback if it should produce an error?

kinyoklion: Before.

sdktests/server_side_hooks.go
```diff
@@ -269,8 +272,58 @@ func beforeEvaluationDataPropagatesToAfterMigration(t *ldtest.T) {
 	})
 }
 
+// This test is meant to check Requirement HOOKS:1.3.7:
+// The client MUST handle exceptions which are thrown (or errors returned, if idiomatic for the language)
+// during the execution of a stage or handler allowing operations to complete unaffected.
+func errorInBeforeStageDoesNotAffectAfterStage(t *ldtest.T) {
+	const numHooks = 100 // why not?
```
kinyoklion, line 279: Because the test output is not decipherable?

tanderson-ld:
I approve, but since we paired on this and others have started reviewing, I will not officially approve to give other reviewers an opportunity before it merges.

sdktests/server_side_hooks.go
tanderson-ld: Ya know, he's got a point.

docs/service_spec.md
```diff
@@ -194,6 +194,9 @@ A `POST` request indicates that the test harness wants to start an instance of t
       * `data` (object, optional): Contains data which should return from different execution stages.
         * `beforeEvaluation` (object, optional): A map of `string` to `ldvalue` items. This should be returned from the `beforeEvaluation` stage of the test hook.
         * `afterEvaluation` (object, optional): A map of `string` to `ldvalue` items. This should be returned from the `afterEvaluation` stage of the test hook.
+      * `errors` (object, optional): Specifies that an error should be returned/exception thrown from a stage. The error message itself is not tested by the framework at this time, as it is not a specified behavior.
+        * `beforeEvaluation` (string, optional): The error/exception message that should be generated in the `beforeEvaluation` stage of the test hook. 
+        * `afterEvaluation` (string, optional): The error/exception message that should be generated in the `afterEvaluation` stage of the test hook.
```
tanderson-ld, line 199: If both data and errors are provided, which should take priority?

sdktests/testapi_hooks.go
```diff
@@ -78,3 +84,28 @@ func (h *Hooks) ExpectCall(t *ldtest.T, hookName string,
...
+	totalCalls := len(hookNames)
+
+	for _, hookName := range hookNames {
+		go func(name string) {
+			out <- helpers.TryReceive(h.instances[name].hookService.CallChannel, hookReceiveTimeout)
+		}(hookName)
+	}
+
+	payloads := make([]servicedef.HookExecutionPayload, totalCalls)
```
kinyoklion, line 101: I think this is incorrect. It makes a slice of the length of expected calls. Then it appends to that list which will then always be longer than the expected calls.

kinyoklion:
```suggestion
	payloads := make([]servicedef.HookExecutionPayload, 0)
```

docs/service_spec.md
cwaldren-ld: It should take the place of posting. So, logic lives before the callback post.

sdktests/server_side_hooks.go
cwaldren-ld: Thank you for providing a reason :D

docs/service_spec.md
cwaldren-ld: I'll update it so error takes priority.

sdktests/testapi_hooks.go
cwaldren-ld: Yea, should have been `make(t, 0, totalCalls)`.

kinyoklion (approved):

PR #204, opened 2024-04-11
feat: Add support for gzipped event payloads

keelerm84:


keelerm84:
Not sure yet what other types of tests we may need, so merging this into a feature branch for now. Once we know for sure the spec isn't changing, I can add any missing coverage and then release this branch.

tanderson-ld (approved):

kinyoklion (approved):

PR #208, opened 2024-04-26
feat: support testing TLS options with two new capabilities

cwaldren-ld:
So far we haven't tested SDKs using https protocol in contract tests. 

This is usually not a problem since each language has some http library that is probably well tested. But, it's a bit more manual in some languages, including C++.

This PR adds two new capabilities. The first is `tls:skip-verify-peer`. If provided, then the SDK under test will receive:
* `skipVerifyPeer bool`: whether or not the peer should be verified

This allows for tests of streaming/polling/events under the https protocol using a self-signed certificate.

The second is `tls:verify-peer`. If provided, the harness will make the SDK under test connect using https and present a self-signed certificate. This should be rejected by the SDK. 

In this initial PR, there is no support for configuring a CA. When I add that support, it will be a 3rd capability that interacts with the other two.


framework/harness/harness.go
```diff
@@ -34,6 +34,7 @@ func NewTestHarness(
 	testServiceBaseURL string,
 	testHarnessExternalHostname string,
 	testHarnessPort int,
+	https bool,
```
cwaldren-ld, line 37: How to tell the harness to use HTTPS. It could go like this:
1. Query the service under test
2. If it has `tls`, then start the HTTPS server 
3. Otherwise start the HTTP server

But this means we can't test both TLS and no TLS, it's just one or the other. 

Could solve this like:

* Run only a subset of tests that are representative. Then run them 2x in both configs.
* Have the test harness run the whole suite twice automatically
* Don't worry about testing only https. That is after all the production configuration. Although, sometimes people talk to Relays without certificates (majority of the time maybe.)
* ??

sdktests/testapi_sdk_client.go
```diff
@@ -159,6 +160,16 @@ func TryNewSDKClient(t *ldtest.T, configurers ...SDKConfigurer) (*SDKClient, err
 			config.ClientSide = o.Some(cs)
 		}
 	}
+	if t.Capabilities().Has(servicedef.CapabilityTLS) {
+		config.TLS = o.Some(servicedef.SDKConfigTLSParams{
+			VerifyPeer: false,
+		})
+		config.ServiceEndpoints = o.Some(servicedef.SDKConfigServiceEndpointsParams{
```
cwaldren-ld, line 167: This is not the correct way to do this.

framework/harness/test_service.go
```diff
@@ -12,27 +12,11 @@ import (
 	"sync"
 	"time"
 
+	"github.com/launchdarkly/sdk-test-harness/v2/serviceinfo"
+
 	"github.com/launchdarkly/sdk-test-harness/v2/framework"
 )
 
-// TestServiceInfo is status information returned by the test service from the initial status query.
```
cwaldren-ld, line 18: This has been refactored out into a new single-purpose `serviceinfo` package because I needed to introduce a circular dependency.

certificate/cert.crt
```diff
@@ -0,0 +1,35 @@
+-----BEGIN CERTIFICATE-----
```
cwaldren-ld: TODO: These certs exist in the filesystem, so during development this all works fine. But when the service binaries are downloaded, they won't exist.

Strategy: probably `go:embed` them in the binary.. then spit them out into temp dirs? Ideally, I could just pass the raw bytes into the `http.ListenAndServeTLS()` but it takes file paths. I've seen some hacks to work around it, but when I tried them, I got odd errors.

framework/harness/certificate/cert.crt
tanderson-ld, line 1: Make cert expiration 10 years or something large.

framework/harness/certificate/gen.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -e
+
+openssl req -new -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -out certificates/cert.cert -keyout certificates/cert.key
```
tanderson-ld: Make cert expiration 10 years or something large.

cwaldren-ld: TODO: increase this to like 5 years?

sdktests/common_tests_base.go
```diff
@@ -109,6 +107,67 @@ func (c commonTestsBase) availableFlagRequestMethods() []flagRequestMethod {
...
+	configurer SDKConfigurer
+}
+
+// Run invokes T.Run() with the protocol's name, passing in a modified T that is suitable for the test.
+func (t transportProtocol) Run(tester *ldtest.T, action func(*ldtest.T)) {
+	// This is a pretty nasty hack. We're modifying the TestHarness that is stashed away in T, in order
+	// to tell it to use HTTPS when creating mock endpoints. This is necessary because higher level
+	// test components - like the mock data sources or event sink - use those methods in their own setup.
+	// So, if this is a test that should use HTTPS, tweak the global TestHarness and enable it - then undo
+	// it after the test runs. WARNING: this won't work with tests that run in parallel.
```
tanderson-ld, line 125: This seems questionable and might be worth some pairing on to try to find a solution that is more elegant.  Do you want to spend some time on Friday thinking through this?

framework/harness/certificate/gen.sh
cwaldren-ld: It's 10 years now

sdktests/common_tests_base.go
cwaldren-ld: Yea, quite possibly. I already explored some other (nastier) solutions but pairing would help

tanderson-ld (approved):

PR #215, opened 2024-05-17
feat: add custom CA capability

cwaldren-ld:
This adds a new capability `tls:custom-ca`. SDKs advertising this capability will be passed a path to a root CA bundle. 

This bundle should be used when creating a TLS connection. The harness uses this capability to test that the SDK can trust a self-signed certificate. 

I've tested this with an implementation in the C++ Client and Server, for polling+streaming+events.


framework/harness/certificate/ca.conf
```diff
@@ -0,0 +1,9 @@
+[req]
```
cwaldren-ld, line 1: TODO: validate that all of these are necessary

docs/service_spec.md
```diff
@@ -223,6 +233,10 @@ A `POST` request indicates that the test harness wants to start an instance of t
         * `afterEvaluation` (string, optional): The error/exception message that should be generated in the `afterEvaluation` stage of the test hook. 
   * `tls` (object, optional): If specified, contains configuration for establishing TLS connections.
     * `skipVerifyPeer` (bool, optional): If true, the SDK's TLS configuration should be set skip verification of the peer. If false or omitted, the SDK should perform peer verification.
+    * `customCAFile` (string, optional): If set, contains a file path pointing to a custom CA file that should be
```
cwaldren-ld, line 236: todo: decide on renaming this `caBundlePath`?

framework/harness/harness.go
```diff
@@ -14,41 +15,58 @@ import (
...
+//go:embed certificate/leaf_private.pem
 var privateKey []byte
 
+//go:embed certificate/ca_public.pem
+var caCertificate []byte
+
 type certPaths struct {
 	cert string
 	key  string
+	ca   string
```
tanderson-ld: caPool or caFile may be a better name if this file can contain more than one CA.  I see ca and think this is something that can issue certs (i.e. has a private key for a ca)

sdktests/common_tests_base.go
```diff
@@ -149,7 +151,19 @@ func (c commonTestsBase) withHTTPSTransportSkipVerifyPeer(t *ldtest.T) transport
...
-	return transportProtocol{"https", configurer}
+	return transportProtocol{"https", "https-skip-verify-peer", configurer}
+}
+
+func (c commonTestsBase) withHTTPSTransportVerifyPeerCustomCA(t *ldtest.T, customCAPath string) transportProtocol {
+	t.RequireCapability(servicedef.CapabilityTLSCustomCA)
+	configurer := helpers.ConfigOptionFunc[servicedef.SDKConfigParams](func(configOut *servicedef.SDKConfigParams) error {
+		configOut.TLS = o.Some(servicedef.SDKConfigTLSParams{
+			SkipVerifyPeer: false,
+			CustomCAFile:   customCAPath,
```
tanderson-ld: file or path?  Pick one?

framework/harness/certificate/leaf_private.pem
tanderson-ld, line 1: Why have any of the harness's keys or certs in source control?  Can they always be generated at runtime?

tanderson-ld (approved):

kinyoklion:
I encountered a couple issues. Setting only the capability of "tls:custom-ca" resulted in configurations that could not pass. Potentially that capability needs to be contingent on verify peer.

The second issue is that http sessions seem to be getting re-used which can cause a test that is intended to fail to pass.  (Resetting SSL state or disabling keep alive in the server resolved the issue.)

I've previously had a problem with keepalive in other SDKs related to performance. (Worked around them with connection close headers.)

framework/harness/certificate/leaf_private.pem
cwaldren-ld: They could be, but it would require:
1) Shelling out to OpenSSL (or some pretty verbose Go code)
2) Not knowing exactly how the certificates look. Since they're committed, if anyone has issues we know exactly what certificates are being passed around.

sdktests/common_tests_base.go
cwaldren-ld: It's a path.. but it points to a file.. I'm stuck in the depths of naming things.. (will use file)

PR #227, opened 2024-08-29
feat: Add filtering-strict capability

keelerm84:


docs/service_spec.md
```diff
@@ -134,6 +134,12 @@ and will send a `?filter=name` query parameter along with streaming/polling requ
 For tests that involve filtering, the test harness will set the `filter` property of the `streaming` or `polling` configuration
 object. The property will either be omitted if no filter is requested, or a non-empty string if requested.
 
+#### Capability `"filtering-strict"`
+
+When initially adding support for the `filtering` capability, the test harness assumed there were no character restrictions on the payload filter key. Later investigation has determined this is not the case.
+
+If this capability is set, the test harness will send a filter key that contains a character that is not allowed in a filter key. The test service should ignore the invalid key and connect as if no filter was specified.
```
cwaldren-ld, line 141: It seems like this should not actually allow the SDK to connect. Otherwise a mis-typed or somehow misconfigured, but still present filter key, will silently behave as if you have no filter key.

keelerm84: The SDKs should log a warning that the provided filter key is invalid. But for the test harness purposes, we can't really check that.

From the SDK perspective, logging and warning but not preventing them from moving forward seems reasonable since we do something similar with context builders.

keelerm84: Given the discussion from yesterday's meeting, I believe we have decided to leave this as is. Going to resolve this comment.

cwaldren-ld (approved):

PR #229, opened 2024-09-05
chore: Update data sources to send FDv2-compatible payloads

keelerm84:
This initial commit is quite limited in scope. It updates the existing server-side contract tests to send the FDv2 equivalent payloads to SDKs under tests.

These changes do not include **any new tests**. Additional commits are required to expand test coverage to include FDv2 specific handling (e.g. xfer-changes, state handling).

It should also be noted these changes have likely broken the client, mobile, PHP, and Roku integrations. That's okay because we will be updating each of those in subsequent commits as well as the FDv2 work progresses.

### DIFF
--- a/framework/types.go
+++ b/framework/types.go
@@ -36,7 +36,8 @@ type Payload struct {
 	Reason string `json:"reason"`
 }
 
-// BaseObject is is the general shape of a put-object event. The delete-object is the same, with the object field being nil.
+// BaseObject is is the general shape of a put-object event. The delete-object
+// is the same, with the object field being nil.
 type BaseObject struct {
 	Version int             `json:"version"`
 	Kind    string          `json:"kind"`
@@ -49,6 +50,7 @@ type PayloadTransferred struct {
 	Version int    `json:"version"`
 }
 
+//nolint:godox
 // TODO: Todd doesn't have this in his spec. What are we going to do here?
 
 type ErrorEvent struct {
--- a/framework/types.go
+++ b/framework/types.go
@@ -16,8 +16,8 @@ type PayloadEvent struct {
 }
 
 type ChangeSet struct {
-	intent *ServerIntent
-	events []eventsource.Event
+	intent *ServerIntent       //nolint:unused
+	events []eventsource.Event //nolint:unused
 }
 
 type ServerIntent struct {
--- a/framework/types.go
+++ b/framework/types.go
@@ -50,7 +50,7 @@ type PayloadTransferred struct {
 }
 
 // TODO: Todd doesn't have this in his spec. What are we going to do here?
-//
+
 //nolint:godox
 type ErrorEvent struct {
 	PayloadID string `json:"payloadId"`
--- a/framework/types.go
+++ b/framework/types.go
@@ -1,0 +1,68 @@
+package framework
+
+import (
+	"encoding/json"
+
+	"github.com/launchdarkly/eventsource"
+)
+
+type PollingPayload struct {
+	Events []PayloadEvent `json:"events"`
+}
+
+type PayloadEvent struct {
+	Name      string      `json:"name"`
+	EventData interface{} `json:"data"`
+}
+
+type ChangeSet struct {
+	intent *ServerIntent
+	events []eventsource.Event
+}
+
+type ServerIntent struct {
+	Payloads []Payload `json:"payloads"`
+}
+
+type Payload struct {
+	// The id here doesn't seem to match the state that is included in the
+	// payload transferred object.
+
+	// It would be nice if we had the same value available in both so we could
+	// use that as the key consistently throughout the process.
+	ID     string `json:"id"`
+	Target int    `json:"target"`
+	Code   string `json:"code"`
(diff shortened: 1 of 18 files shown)


framework/types.go
```diff
@@ -0,0 +1,68 @@
+package framework
```
keelerm84, line 1: Same basic types I'm carrying around everywhere right now. Need to get these into the events package or somewhere core.

mockld/polling_service.go
```diff
@@ -152,29 +152,84 @@ func (p *PollingService) pollingHandler(getDataFn func(*PollingService, *http.Re
 
 func (p *PollingService) standardPollingHandler() http.Handler {
 	return p.pollingHandler(func(p *PollingService, r *http.Request) []byte {
-		return p.currentData.Serialize()
+		fdv2SdkData, ok := p.currentData.(FDv2SDKData)
```
keelerm84, line 155: This translation bit is a mess and is going to require the most rework. I wanted to see how the changes emerged as we developed additional FDv2 tests (like the xfer-changes stuff) before I tried to fix this hack too much.

mockld/polling_service.go
```diff
@@ -152,29 +152,84 @@ func (p *PollingService) pollingHandler(getDataFn func(*PollingService, *http.Re
...
+
+		return data
 	})
 }
 
 func (p *PollingService) phpFlagHandler() http.Handler {
 	return p.pollingHandler(func(p *PollingService, r *http.Request) []byte {
-		data, _ := p.currentData.(ServerSDKData)
-		return data["flags"][mux.Vars(r)["key"]]
+		// TODO: Update this logic
```
keelerm84, line 213: PHP stuff will come later.

mockld/polling_service_test.go
```diff
@@ -1,86 +1,84 @@
...
-	doPollingServiceTests(
-		t,
-		ServerSideSDK,
-		EmptyServerSDKData(),
-		NewServerSDKDataBuilder().RawFlag("flag1", json.RawMessage(`{"key": "flag1"}`)).Build(),
-		"GET",
-		"/sdk/latest-all",
-	)
-}
+// TODO: Re-enable these tests
```
keelerm84, line 18: I will fix the unit tests later. Again, just getting something working out in front of you.

mockld/sdk_data.go
```diff
@@ -41,6 +43,12 @@ type SDKData interface {
 	Serialize() []byte
 }
 
+type FDv2SDKData []framework.BaseObject
```
keelerm84, line 46: This type may have to be more complex over time. We might need something to control the state, the version, and potential the server intent. That way the serialize method can render the full payload and we don't have to do that hacky translation I called out earlier.

mockld/sdk_data.go
```diff
@@ -63,6 +71,23 @@ func (b blockingUnavailableSDKData) Serialize() []byte { return nil }
 // We use this for both regular server-side SDKs and the PHP SDK.
 type ServerSDKData map[DataItemKind]map[string]json.RawMessage
 
+func (s ServerSDKData) AsFDv2SDKData(t *ldtest.T) FDv2SDKData {
```
keelerm84: The data model in the parameterized tests used to match the SDK format. That doesn't work anymore. For now, I do a simple translation. Ideally we would do this when loading the files.

keelerm84 (commented):
This is a VERY rough first pass at getting the test harness working with FDv2. Lots of todos, but I figure it is better to start getting stuff reviewed and merged so it is accessible to both of us.

sdktests/php_eval.go
```diff
@@ -10,9 +10,9 @@ func doPHPEvalTests(t *ldtest.T) {
 	// how we set up the polling endpoints in the mock data source, and that is handled
 	// transparently by the logic in mockld/polling_service.go based on the fact that the
 	// "SDK kind" configured for test suite is PHP.
-	t.Run("parameterized", runParameterizedServerSideEvalTests)
```
cwaldren-ld: Do these deserve TODOs? I feel like it could get lost.

sdktests/testapi_sdk_data.go
```diff
@@ -95,7 +100,9 @@ func NewSDKDataSourceWithoutEndpoint(t *ldtest.T, data mockld.SDKData, options .
 		d.streamingService = mockld.NewStreamingService(data, sdkKind, t.DebugLogger())
 	}
 
-	t.Debug("setting SDK data to: %s", string(data.Serialize()))
+	jsonStr, _ := json.Marshal(data)
```
cwaldren-ld: could check the error, then use `t` to assert it is nil

cwaldren-ld (approved):

sdktests/php_eval.go
keelerm84: I can actually add this back in, so we are good there.

sdktests/testapi_sdk_data.go
keelerm84: Updated!

PR #230, opened 2024-09-05
feat: Add tests which evaluate complex JSON variations.

kinyoklion:


data/data-files/server-side-eval/json-variations.yml
```diff
@@ -0,0 +1,43 @@
+---
+name: JSON variations
+
+constants:
+  CONTEXT: { kind: "user", key: "x"}
+
+parameters:
+- TYPE_NAME: any
+  TYPE_VALUE_1: ["a", true, null, 3.14, {str: "str", double: 0.5, 'null': null}]
+  TYPE_VALUE_2: {str: "str", double: 0.5, 'null': null, arr: ["a", true, null, 3.14, {str: "str", double: 0.5, 'null': null}]}
```
kinyoklion, line 10: Node server SDK was not handling null in JSON variations correctly. It doesn't seem there are any existing tests which verify JSON variations sufficiently. Instead they use raw types, which our UI doesn't allow for JSON variations.

cwaldren-ld (approved):

PR #238, opened 2024-09-23
chore: Add new, optional top level SDK data system configuration

keelerm84:


servicedef/sdk_config.go
```diff
@@ -104,3 +105,47 @@ type SDKConfigWrapper struct {
 	Name    string `json:"name"`
 	Version string `json:"version"`
 }
+
+type SDKConfigDataSystem struct {
+	Synchronizers []SDKConfigDataSystemDataSource `json:"synchronizers,omitempty"`
```
keelerm84, line 110: @cwaldren-ld I know we talked about only having a primary and secondary. It seems reasonable to me to model it as an array for future expansion, but enforce that order by convention. It also allows opting in to secondary support nice as it doesn't really change the shape. You just process the next item on the list in order.

servicedef/sdk_config.go
```diff
@@ -104,3 +105,47 @@ type SDKConfigWrapper struct {
...
+
+type DataSystemDataSourceConfig interface{}
+
+type DataSystemStreamingDataSourceConfig struct {
+	SDKConfigStreamingParams
+}
+
+type DataSystemPollingDataSourceConfig struct {
+	SDKConfigPollingParams
+}
```
keelerm84: I wasn't sure the best way to do this in this language, so figured I should just ask you @cwaldren-ld.

Basically we want to have the 2 separate types of configs (streaming vs polling), but I also want a key telling me which type it is.

How would you structure this?

servicedef/sdk_config.go
```diff
@@ -104,3 +105,47 @@ type SDKConfigWrapper struct {
...
+
+type SDKConfigDataSystemDataSource struct {
+	Type string `json:"type"`
+	DataSystemDataSourceConfig
+}
+
+type DataSystemDataSourceConfig interface{}
+
+type DataSystemStreamingDataSourceConfig struct {
+	SDKConfigStreamingParams
```
keelerm84: Embedding the existing polling and streaming configs we are defining and using above.

servicedef/sdk_config.go
cwaldren-ld: Let's go with this and see if the "shape" works fine for SDks that implement the test.

servicedef/sdk_config.go
```diff
@@ -104,3 +105,55 @@ type SDKConfigWrapper struct {
...
+	Initializers  []SDKConfigDataSystemDataSource `json:"initializers,omitempty"`
+	Persistence   SDKConfigDataSystemPersistence  `json:"persistence,omitempty"`
+}
+
+type SDKConfigDataSystemDataSource struct {
+	Type string `json:"type"`
+	DataSystemDataSourceConfig
+}
+
+type DataSystemDataSourceConfig interface{}
```
cwaldren-ld, line 120: What is this line doing?

servicedef/sdk_config.go
```diff
@@ -104,3 +105,55 @@ type SDKConfigWrapper struct {
...
+
+type SDKConfigDataSystemPersistence struct {
+	Store SDKConfigDataSystemPersistenceStore `json:"store"`
+	Cache SDKConfigDataSystemPersistenceCache `json:"cache"`
+}
+
+type SDKConfigDataSystemPersistenceType string
+
+const (
+	Redis    SDKConfigDataSystemPersistenceType = "redis"
```
cwaldren-ld: Another style of doing this is `Redis = SDKConfigDataSystemPersistenceType("redis")`

I think they are equivalent.

servicedef/sdk_config.go
```diff
@@ -104,3 +105,55 @@ type SDKConfigWrapper struct {
...
+
+const (
+	Off      SDKConfigDataSystemPersistenceMode = "off"
+	TTL      SDKConfigDataSystemPersistenceMode = "ttl"
+	Infinite SDKConfigDataSystemPersistenceMode = "infinite"
+)
+
+type SDKConfigDataSystemPersistenceCache struct {
+	Mode string `json:"mode"`
+	TTL  int    `json:"ttl"` // This value is only valid when the Mode is set to TTL. It must be a positive integer.
```
cwaldren-ld: I think this could be `o.Maybe[int]`, with tag `json:"ttl,omitempty`

servicedef/sdk_config.go
cwaldren-ld: I think the way we've handled this before is two different fields.

```golang
type Thing struct {
       Type string `json:"type"`
       OptionA o.Maybe[OptionA] `json:"optionA,omitempty"`
       OptionB o.Maybe[OptionB] `json:"optionB,omitempty"`
}
```

For example, our Command config (just using C++ for convenience on my part)
```
struct CommandParams {
    Command command; <-- enum
    std::optional<EvaluateFlagParams> evaluate;
    std::optional<EvaluateAllFlagParams> evaluateAll;
    std::optional<CustomEventParams> customEvent;
    std::optional<IdentifyEventParams> identifyEvent;
    std::optional<ContextBuildParams> contextBuild;
    std::optional<ContextConvertParams> contextConvert;
    CommandParams();
};
```

keelerm84: Updated to match your suggested format.

servicedef/sdk_config.go
keelerm84: Updated.

servicedef/sdk_config.go
keelerm84: Done.

cwaldren-ld (approved):

PR #239, opened 2024-10-04
chore: Introduction of redis-backed daemon-mode tests

keelerm84:


sdktests/server_side_data_system_base.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
+	"github.com/launchdarkly/go-server-sdk-evaluation/v3/ldbuilders"
+	"github.com/launchdarkly/sdk-test-harness/v2/framework/ldtest"
+	"github.com/launchdarkly/sdk-test-harness/v2/servicedef"
+	"github.com/redis/go-redis/v9"
+	"github.com/stretchr/testify/require"
+)
+
+func doServerSideDataSystemTests(t *ldtest.T) {
+	rdb := redis.NewClient(&redis.Options{
```
keelerm84: A couple things to note. As this work currently stands, this only supports redis. Since this work is being merged into a separate feature branch, I'm taking a very iterative approach here.

In this first pass, we just expect that you have redis up and running at this host and port. As the tests expand, that will eventually include work to support dynamodb and consul. At that point, we will probably start refactoring out into some docker container stuff.

All that to say, I know this isn't ideal, but a lot of this is a stepping stone.

sdktests/server_side_data_system_base.go
```diff
@@ -0,0 +1,112 @@
...
+
+	uncachedFlagKeyBytes, err :=
+		ldbuilders.NewFlagBuilder("uncached-flag-key").Version(1).
+			On(true).Variations(ldvalue.String("off"), ldvalue.String("match"), ldvalue.String("fallthrough")).
+			OffVariation(0).
+			FallthroughVariation(2).
+			Build().MarshalJSON()
+	require.NoError(t, err)
+
+	initialFlags["uncached-flag-key"] = string(uncachedFlagKeyBytes)
```
keelerm84: Just some initial data to play with so I don't have to define it every single time.

sdktests/server_side_data_system_base.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (s *ServerSideDataSystemTests) Run(t *ldtest.T) {
+	t.Run("uses default prefix", s.usesDefaultPrefix)
+	t.Run("uses custom prefix", s.usesCustomPrefix)
+
+	t.Run("read-only", s.doReadOnlyTests)
+}
+
+func (s *ServerSideDataSystemTests) usesDefaultPrefix(t *ldtest.T) {
+	require.NoError(t, s.persistence.Reset())
```
keelerm84: Right now the redis implementation defines Reset as a FlushAll, which clears everything. That makes it super fast. Depending on later needs, that might be restarting a docker container, or whatever is needed for other integrations.

sdktests/server_side_data_system_persistence_store.go
```diff
@@ -0,0 +1,35 @@
+package sdktests
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/redis/go-redis/v9"
+)
+
+type PersistenceStore interface {
```
keelerm84: Very rough first pass. Will have to see what need arise.

sdktests/server_side_data_system_read_only.go
```diff
@@ -0,0 +1,185 @@
...
+	o "github.com/launchdarkly/sdk-test-harness/v2/framework/opt"
+	"github.com/launchdarkly/sdk-test-harness/v2/servicedef"
+	"github.com/stretchr/testify/require"
+)
+
+func (s *ServerSideDataSystemTests) doReadOnlyTests(t *ldtest.T) {
+	t.Run("daemon mode", s.daemonModeTests)
+}
+
+func (s *ServerSideDataSystemTests) daemonModeTests(t *ldtest.T) {
```
keelerm84: Interesting tests finally start!

sdktests/testapi_sdk_data_system.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"errors"
+
+	"github.com/launchdarkly/sdk-test-harness/v2/framework/opt"
+	o "github.com/launchdarkly/sdk-test-harness/v2/framework/opt"
+	"github.com/launchdarkly/sdk-test-harness/v2/servicedef"
+)
+
+type DataSystem struct {
```
keelerm84: Again, I think we are going to see some different patterns we may want to emerge as we build out more tests. Just KISSing it for now.

sdktests/server_side_data_system_base.go
```diff
@@ -0,0 +1,116 @@
...
+		},
+	})
+
+	client := NewSDKClient(t, dataSystem)
+
+	require.Never(
+		t,
+		checkForUpdatedValue(t, client, "flag-key", ldcontext.New("user-key"),
+			ldvalue.String("default"), ldvalue.String("fallthrough"), ldvalue.String("default")),
+		time.Millisecond*100,
```
cwaldren-ld: Let's pull these into named constants

sdktests/server_side_data_system_persistence_store.go
```diff
@@ -0,0 +1,35 @@
...
+
+type RedisPersistenceStore struct {
+	redis *redis.Client
+}
+
+func (r RedisPersistenceStore) DSN() string {
+	return fmt.Sprintf("redis://%s", r.redis.Options().Addr)
+}
+
+func (r RedisPersistenceStore) Reset() error {
```
cwaldren-ld: Since these methods are mutating, I think it'd be slightly clearer to have them operate on pointer receivers.

sdktests/server_side_data_system_persistence_store.go
cwaldren-ld: Since we call this `PersistentStore` elsewhere, should we call it that here? or is the concept slightly different?

cwaldren-ld: If possible, prefer to define the interface near where it is used (rather than in the place that implements it)

sdktests/server_side_data_system_read_only.go
```diff
@@ -0,0 +1,195 @@
...
+			FlagKey:      "flag-key",
+			Context:      o.Some(context),
+			ValueType:    servicedef.ValueTypeAny,
+			DefaultValue: ldvalue.String("default"),
+			Detail:       true,
+		})
+
+		return result.Reason.IsDefined() &&
+			result.Reason.Value().GetErrorKind() == ldreason.EvalErrorFlagNotFound
+	}, time.Second, time.Millisecond*20, "flag was found before it should have been")
```
cwaldren-ld: Same comment re: pulling into a constant to keep these consistent

cwaldren-ld (commented):
Really excited about this. No joke.

sdktests/server_side_data_system_persistence_store.go
keelerm84: I struggled with this and would gladly accept some suggestions.

Basically the persistence store here is just the way for the test harness to update redis, consul, dynamodb. It is conceptually different from the SDK's concept of a persistence store in that the methods and expectations around behavior are very different.

> If possible, prefer to define the interface near where it is used (rather than in the place that implements it)

I'm not sure how that would differ much since the interface and implementation are being defined in the same package that it is used. There isn't really a single file that is going to use either of those, unless I just throw all the persistence tests into one file.

Or maybe I'm just misunderstanding the request.

sdktests/server_side_data_system_read_only.go
keelerm84: These parameters are used throughout the code base anywhere the eventually or never methods are used.

Do you want them to be global constants somewhere then? And if so, can I do that in a separate PR so this one doesn't become very noisy with those changes?

sdktests/server_side_data_system_persistence_store.go
keelerm84: Done.

sdktests/server_side_data_system_persistence_store.go
cwaldren-ld: The requirements of that interface are defined by the thing that uses it, not the thing that implements it. So the ask would just be that the interface is defined close by the consumer.

E.g.
```golang
type PersistenceStore {
   ... defines the requirements that TestExecutor has
}

type TestExecutor struct {
      store PersistenceStore
}
```

This way, you could implement the different consule/redis/dynamoDB wherever you want, they don't need to all be in the same package. It also stops the proliferation of a bunch of consumers sharing an overly broad interface, where they should really be defining their own.

cwaldren-ld: Since it's all in the same package (I think), some of these benefits aren't as clear. Still, having it defined physically right next to the consumer is really helpful.

sdktests/server_side_data_system_read_only.go
cwaldren-ld: Oh 😢 

Maybe just go with the flow then.

sdktests/server_side_persistence_base.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (s *ServerSidePersistenceTests) Run(t *ldtest.T) {
+	t.Run("uses default prefix", s.usesDefaultPrefix)
+	t.Run("uses custom prefix", s.usesCustomPrefix)
+
+	t.Run("daemon mode", s.doDaemonModeTests)
+}
+
+func (s *ServerSidePersistenceTests) usesDefaultPrefix(t *ldtest.T) {
+	require.NoError(t, s.persistence.Reset())
```
cwaldren-ld: I think adding _just one_ level of wrapping (perhaps in `(s *ServerSidePersistenceTests) Run(t *ldtest.T)`)? 

could be useful: ensuring test setup/teardown happens the same for every test. Specifically, that `Reset` is called.

Maybe not, I could probably go either way. Also good to call out that these cannot run in parallel.

sdktests/server_side_persistence_daemon.go
```diff
@@ -0,0 +1,184 @@
...
+		Type: servicedef.Redis,
+		DSN:  s.persistence.DSN(),
+	})
+	persistence.SetCache(servicedef.SDKConfigPersistenceCache{
+		Mode: servicedef.TTL,
+		TTL:  o.Some(1),
+	})
+	context := ldcontext.New("user-key")
+
+	t.Run("will cache found flag for TTL", func(t *ldtest.T) {
```
cwaldren-ld: I get it now, but the test name is confusing at first - ".. so, will the cache found this flag..?" 

Although once I got it, I couldn't "unsee" it without trying hard.

How about:
> cache hit persists for TTL

> cache miss persists for TTL

sdktests/server_side_persistence_base.go
cwaldren-ld: One way we could solve it without wrapping/nesting/defer:

Make the API be like:
```
persistence := s.persistence.Open()

persistence.WriteData("foo", "bar")
```

where Open just calls Reset() and then returns the object (and open is the only function on `s.persistence` or whatever.)

sdktests/server_side_persistence_base.go
```diff
@@ -0,0 +1,112 @@
...
+		checkForUpdatedValue(t, client, "flag-key", ldcontext.New("user-key"),
+			ldvalue.String("default"), ldvalue.String("fallthrough"), ldvalue.String("default")),
+		time.Millisecond*100,
+		time.Millisecond*20,
+		"flag value was updated, but it should not have been",
+	)
+
+	require.NoError(t, s.persistence.WriteData(customPrefix+":features", s.initialFlags))
+
+	pollUntilFlagValueUpdated(t, client, "flag-key", ldcontext.New("user-key"),
```
cwaldren-ld, line 118: is the `pollUntil` an assertion..? It's odd that it is so incongruent with the `requires`.

sdktests/testapi_sdk_persistence.go
```diff
@@ -0,0 +1,38 @@
...
+	p.Store = o.Some(store)
+}
+
+func (p *Persistence) SetCache(cache servicedef.SDKConfigPersistenceCache) {
+	p.Cache = o.Some(cache)
+}
+
+func (p Persistence) Configure(target *servicedef.SDKConfigParams) error {
+	if !p.Store.IsDefined() || !p.Cache.IsDefined() {
+		return errors.New("Persistence must have a store configuration")
```
cwaldren-ld:
```suggestion
		return errors.New("Persistence must have a store and cache configuration")
```

servicedef/sdk_config.go
```diff
@@ -26,7 +26,7 @@ type SDKConfigParams struct {
 	ClientSide          o.Maybe[SDKConfigClientSideParams]          `json:"clientSide,omitempty"`
 	Hooks               o.Maybe[SDKConfigHooksParams]               `json:"hooks,omitempty"`
 	Wrapper             o.Maybe[SDKConfigWrapper]                   `json:"wrapper,omitempty"`
-	DataSystem          o.Maybe[SDKConfigDataSystem]                `json:"dataSystem,omitempty"`
+	Persistence         o.Maybe[SDKConfigPersistence]               `json:"persistence,omitempty"`
```
cwaldren-ld: I just noticed the `PersistentDataStore` config in this struct. Hmm. We could just use that I guess? it seems to not be used.

sdktests/server_side_persistence_daemon.go
keelerm84: [Will It Cache?](https://www.youtube.com/watch?v=lAl28d6tbko)

sdktests/server_side_persistence_base.go
keelerm84: Yes. It is a wrapper for `RequiresEventually`.

sdktests/server_side_persistence_base.go
keelerm84: I expect this interface is going to change more as I add the other database integrations, and do the read-write types. 

Mind if I leave this as is, and then in a separate task, you and I both circle back together and see what kind of interface we can shake out?

servicedef/sdk_config.go
keelerm84: I will de-dupe as needed.

sdktests/server_side_persistence_base.go
cwaldren-ld: Yup sounds good

sdktests/server_side_persistence_base.go
cwaldren-ld: ah. I guess another task, but we should rename it to requireFlagValueUpdatesEventually or something

servicedef/sdk_config.go
```diff
@@ -12,21 +12,20 @@ import (
...
+	TLS                  o.Maybe[SDKConfigTLSParams]                  `json:"tls,omitempty"`
+	Streaming            o.Maybe[SDKConfigStreamingParams]            `json:"streaming,omitempty"`
+	Polling              o.Maybe[SDKConfigPollingParams]              `json:"polling,omitempty"`
+	Events               o.Maybe[SDKConfigEventParams]                `json:"events,omitempty"`
+	BigSegments          o.Maybe[SDKConfigBigSegmentsParams]          `json:"bigSegments,omitempty"`
+	Tags                 o.Maybe[SDKConfigTagsParams]                 `json:"tags,omitempty"`
+	ClientSide           o.Maybe[SDKConfigClientSideParams]           `json:"clientSide,omitempty"`
+	Hooks                o.Maybe[SDKConfigHooksParams]                `json:"hooks,omitempty"`
+	Wrapper              o.Maybe[SDKConfigWrapper]                    `json:"wrapper,omitempty"`
+	PersistenceDataStore o.Maybe[SDKConfigPersistenceDataStoreParams] `json:"persistenceDataStore,omitempty"`
```
cwaldren-ld: Ok just one nit. Can we have all the external, SDK facing interfaces use [`PersistentDataStore`](https://docs.launchdarkly.com/sdk/concepts/data-stores) as the terminology. 

I feel like the inconsistency of persistence vs persistent will be one more thing for people to be confused about.

cwaldren-ld (approved):

servicedef/sdk_config.go
keelerm84: Changed it everywhere publicly as requested.

PR #242, opened 2024-10-09
feat: Add `client-prereq-events` capability

keelerm84:


sdktests/client_side_events_summary.go
```diff
@@ -276,3 +282,147 @@ func doClientSideSummaryEventResetTest(t *ldtest.T) {
...
+		Flag(prereq2Key, prereq2Result).
+		Flag(prereq3Key, prereq3Result)
+
+	dataSource := NewSDKDataSource(t, dataBuilder.Build())
+	events := NewSDKEventSinkWithGzip(t, t.Capabilities().Has(servicedef.CapabilityEventGzip))
+	client := NewSDKClient(t,
+		WithClientSideInitialContext(contextA),
+		dataSource, events)
+
+	_ = client.EvaluateFlag(t, servicedef.EvaluateFlagParams{FlagKey: prereq1Key, DefaultValue: default1})
```
keelerm84, line 338: Okay @kinyoklion check out this setup I have. 

- I explicitly evaluate prereq1Key. This gets added to the summary event, including the default value I provided.
- I then evaluate top level, which has a prereq on prereqs 1 and 2. This results in prereq2 getting added to the summary event without a default value.
- We then evaluate prereq 2 explicitly, which just increments that value, but does not add the default value to the summary counter.

sdktests/client_side_events_summary.go
```diff
@@ -276,3 +282,147 @@ func doClientSideSummaryEventResetTest(t *ldtest.T) {
...
+}
+
+func doClientSideSummaryPrereqUnknownFlagTest(t *ldtest.T) {
+	topLevelKey := "flag1"
+	topLevelResult := mockld.ClientSDKFlag{
+		Value:         ldvalue.String("value1-a"),
+		Variation:     o.Some(0),
+		FlagVersion:   o.Some(1),
+		Version:       11,
+		Prerequisites: []string{"unknown"},
```
keelerm84, line 387: And yet here we have a prereq on an unknown flag. Once we explicitly evaluate the top level key here, the unknown flag is added to the summary event WITH a default value. What's up with that?

kinyoklion (approved):

PR #243, opened 2024-10-11
chore: Introduce read-write persistent data store tests

keelerm84:


sdktests/server_side_persistence_base.go
```diff
@@ -14,52 +14,62 @@ import (
...
+
+		t.Run("redis", newServerSidePersistentTests(t, &RedisPersistentStore{redis: rdb}).Run)
+	}
 }
 
 type PersistentStore interface {
 	DSN() string
 
+	// TODO: Change these names to something less terrible, or make them even
+	// more generic.
```
cwaldren-ld: Since our stores are k/v, perhaps just `Get` and `Set` would be the most familiar names.

sdktests/server_side_persistence_read_write.go
```diff
@@ -0,0 +1,469 @@
...
+func (s *ServerSidePersistentTests) initializesStoreWhenDataReceived(t *ldtest.T) {
+	require.NoError(t, s.persistentStore.Reset())
+
+	persistence := NewPersistence()
+	persistence.SetStore(servicedef.SDKConfigPersistentStore{
+		Type: s.persistentStore.Type(),
+		DSN:  s.persistentStore.DSN(),
+	})
+	persistence.SetCache(servicedef.SDKConfigPersistentCache{
+		Mode: servicedef.Off,
```
cwaldren-ld: I'm wondering if we'll quickly run into some other constant named `Off`. Might be worth prefixing it with `CacheMode` so `CacheModeOff`.

sdktests/server_side_persistence_read_write.go
```diff
@@ -0,0 +1,469 @@
...
+	})
+
+	sdkData := s.makeSDKDataWithFlag("flag-key", 1, ldvalue.String("value"))
+	stream, configurers := s.setupDataSources(t, sdkData)
+	configurers = append(configurers, persistence)
+
+	_, err := s.persistentStore.ReadField("launchdarkly:$inited")
+	require.Error(t, err) // should not exist
```
cwaldren-ld: Is there any ambiguity that this error could be something else besides `not exist`? Like, the connection to the DB is down? 

Basically wondering if the assert here should check the actual error type, more than existence. Or if that's irrelevant.

sdktests/server_side_persistence_read_write.go
```diff
@@ -0,0 +1,469 @@
...
+	})
+
+	sdkData := s.makeSDKDataWithFlag("flag-key", 1, ldvalue.String("value"))
+	stream, configurers := s.setupDataSources(t, sdkData)
+	configurers = append(configurers, persistence)
+
+	_ = NewSDKClient(t, s.baseSDKConfigurationPlus(configurers...)...)
+	s.eventuallyRequireDataStoreInit(t, "launchdarkly")
+
+	require.NoError(t, s.persistentStore.WriteData("launchdarkly:features", s.initialFlags))
```
cwaldren-ld: I think the relationship between the flag versions could be made clearer here. We don't see the setup of `initialFlags` in the test, but we are expecting that `1` is lower and `200` is higher. 

Could the lower/higher be derived from the initial data, or could the initial data be generated by a function call that passes in versions?

sdktests/server_side_persistence_read_write.go
```diff
@@ -0,0 +1,469 @@
...
+
+	// Lower versioned updates are ignored
+	updateData := s.makeFlagData("flag-key", 1, ldvalue.String("new-value"))
+	stream.StreamingService().PushUpdate("flags", "flag-key", updateData)
+	s.neverValidateFlagData(t, "launchdarkly", map[string]m.Matcher{
+		"flag-key":          basicFlagValidationMatcher("flag-key", 1, "new-value"),
+		"uncached-flag-key": basicFlagValidationMatcher("uncached-flag-key", 100, "value"),
+	})
+
+	// Higher versioned updates are applied
```
cwaldren-ld, line 147: We could have one more test for "equal version" (unless that's sort of implicitly being handled by the `uncached-flag-key`?)

sdktests/server_side_persistence_read_write.go
```diff
@@ -0,0 +1,469 @@
...
+			checkForUpdatedValue(t, client, "uncached-flag-key", context,
+				ldvalue.String("default"), ldvalue.String("fallthrough"), ldvalue.String("default")),
+			time.Millisecond*500, time.Millisecond*20, "uncached-flag-key was incorrectly cached")
+	} else if cacheConfig.Mode == servicedef.TTL {
+		// But eventually, it will expire and then we will fetch it from the database.
+		h.RequireEventually(t,
+			checkForUpdatedValue(t, client, "flag-key", context,
+				ldvalue.String("value"), ldvalue.String("fallthrough"), ldvalue.String("default")),
+			time.Second, time.Millisecond*20, "flag-key was incorrectly cached")
+	}
```
cwaldren-ld, line 241: Does the assertion on `uncached-flag-key` need to be in this test, or would it be clearer to have this test operate only on `flag-key`? I don't have a strong opinion.

sdktests/server_side_persistence_read_write.go
```diff
@@ -0,0 +1,469 @@
...
+	pollUntilFlagValueUpdated(t, client, "flag-key", context,
+		ldvalue.String("default"), ldvalue.String("value"), ldvalue.String("default"))
+
+	require.NoError(t, s.persistentStore.WriteData("launchdarkly:features", s.initialFlags))
+
+	// This key was already cached, so it shouldn't see the change above.
+	h.RequireNever(t,
+		checkForUpdatedValue(t, client, "flag-key", context,
+			ldvalue.String("value"), ldvalue.String("new-value"), ldvalue.String("default")),
+		time.Millisecond*500, time.Millisecond*20, "flag-key was incorrectly updated")
```
cwaldren-ld: These time-based assertions must depend on the TTL value to be correct right? Should they be derived from it, like cacheConfig.TTL + X milliseconds or something?

sdktests/server_side_persistence_base.go
keelerm84: Updated.

sdktests/server_side_persistence_read_write.go
keelerm84: Updated.

sdktests/server_side_persistence_read_write.go
keelerm84: Updated to check for the redis case. It will evolve more I think as we introduce those new types.

sdktests/server_side_persistence_read_write.go
keelerm84: Hmmm. I will make a note to circle back around to fix this in a subsequent commit.

I agree the relationship isn't clear, and this is something I will improve before this feature branch is released.

sdktests/server_side_persistence_read_write.go
keelerm84: Added.

sdktests/server_side_persistence_read_write.go
keelerm84: It handles infinite and timed, so the code has some branching. I'll update it and you can let me know if you like this better, or how you want to handle it.

sdktests/server_side_persistence_read_write.go
keelerm84: That assertion is showing that if the flag hasn't been evaluated, and was written to the store after the SDK inits, that the SDK will still see it immediately without having to wait for the cache timeout. So I think having it here is important.

cwaldren-ld (approved):
Approved, pending compiler errors.

PR #248, opened 2024-10-22
chore: Add consul and dynamodb support

keelerm84:


sdktests/server_side_persistence_consul.go
```diff
@@ -0,0 +1,121 @@
...
+
+	consul "github.com/hashicorp/consul/api"
+	"github.com/launchdarkly/sdk-test-harness/v2/servicedef"
+)
+
+type ConsulPersistentStore struct {
+	consul *consul.Client
+}
+
+// {{{ PersistentStore implementation
```
kinyoklion: Is this a fold marker? Is it intentional?

sdktests/server_side_persistence_dynamodb.go
```diff
@@ -0,0 +1,246 @@
...
+	for len(requests) > 0 {
+		batchSize := int(math.Min(float64(len(requests)), 25))
+		batch := requests[:batchSize]
+		requests = requests[batchSize:]
+
+		_, err := client.BatchWriteItem(&dynamodb.BatchWriteItemInput{
+			RequestItems: map[string][]*dynamodb.WriteRequest{table: batch},
+		})
+		if err != nil {
+			// COVERAGE: can't simulate this condition in unit tests because we will only get this
```
kinyoklion: Maybe this comment does apply, but the coverage suppression probably doesn't.

(Though I am not 100% sure it makes sense in its original location, which was also a test.)

sdktests/server_side_persistence_base.go
```diff
@@ -6,6 +6,13 @@ import (
 	"github.com/redis/go-redis/v9"
 	"github.com/stretchr/testify/require"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	c "github.com/hashicorp/consul/api"
```
cwaldren-ld: nit: `c` -> `consul`

sdktests/server_side_persistence_base.go
```diff
@@ -21,16 +28,49 @@ func doServerSidePersistentTests(t *ldtest.T) {
...
+				WithCredentials(
+					credentials.NewStaticCredentials(
+						"dummy",
+						"dummy",
+						"dummy",
+					),
+				),
+		))
+
+		// Create a DynamoDB client from just a session.
```
cwaldren-ld: what is the alternative?

sdktests/server_side_persistence_consul.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (c ConsulPersistentStore) DSN() string {
```
cwaldren-ld: It's not necessarily bad to have this be a value receiver, but it does feel a bit confusing when the rest are pointer receivers. Go syntax sugar will make it work at the callsite. 

It will make a copy of the `ConsulPersistentStore` struct, which could have unintended side-effects (like perhaps there is some mutex deep in the consul client that cannot be copied). Go is great.

sdktests/server_side_persistence_base.go
```diff
@@ -21,16 +28,49 @@ func doServerSidePersistentTests(t *ldtest.T) {
...
 	}
 }
 
 type PersistentStore interface {
 	DSN() string
 
-	Get(key string) (string, error)
-	GetMap(key string) (map[string]string, error)
-	WriteMap(key string, data map[string]string) error
+	Get(prefix, key string) (string, bool, error)
```
cwaldren-ld: It's not immediately clear to me what (string, bool, error) means, but I suspect it's `value, present, error`? 

Doc comment would work, or giving each return value a name would also be fine

cwaldren-ld: Another option: since we hve the whole `Opt` thing going on, this could be an `OptString, error`

sdktests/server_side_persistence_consul.go
```diff
@@ -0,0 +1,121 @@
...
+	return string(pair.Value), true, nil
+}
+
+func (c *ConsulPersistentStore) GetMap(prefix, key string) (map[string]string, error) {
+	kv := c.consul.KV()
+	pairs, _, err := kv.List(prefix+"/"+key, nil)
+
+	if err != nil {
+		return nil, fmt.Errorf("list failed for %s: %s", key, err)
+	}
```
cwaldren-ld, line 49: For errors returned from these wrapper methods, it could make sense to prefix them with the impl name so that it's clear exactly where the error is coming from. I don't have a good suggestion for the prefix though. 
```go
return nil, fmt.Errorf("consul-wrapper: foo bar")
```
ideally Consul itself is prefixing stuff with `consul` as well, but maybe not? 

Because this can get repetitive, there are usually two options:
1. Make a helper like `return nil, c.newErr(...)`
2. Make a dedicated error type like `return nil, &consulErr { action: "list", key: key, err: err }` and have it render the error string


----------------

What you have is also probably fine. This is more of a best practice. Feel free to disregard.

sdktests/server_side_persistence_consul.go
```diff
@@ -0,0 +1,121 @@
...
+		flagKey := strings.TrimPrefix(pair.Key, prefix+"/"+key+"/")
+		results[flagKey] = string(pair.Value)
+	}
+	return results, nil
+}
+
+func (c *ConsulPersistentStore) WriteMap(prefix, key string, data map[string]string) error {
+	kv := c.consul.KV()
+
+	// Start by reading the existing keys; we will later delete any of these that weren't in allData.
```
cwaldren-ld: > allData

think the comment needs to be tweaked

sdktests/server_side_persistence_consul.go
```diff
@@ -0,0 +1,121 @@
...
+	pairs, _, err := kv.List(prefix, nil)
+	if err != nil {
+		return fmt.Errorf("failed to get existing items prior to Init: %s", err)
+	}
+	oldKeys := make(map[string]bool)
```
cwaldren-ld: If `oldKeys` is meant to represent a set, I find it clearer to use `map[string]struct{}`. That way, there's no need to check `v` (on line 81 of this diff).

On line 77, you'd `delete` from the set.

sdktests/server_side_persistence_consul.go
```diff
@@ -0,0 +1,121 @@
...
+			op := &consul.KVTxnOp{Verb: consul.KVDelete, Key: prefix + "/" + key + "/" + k}
+			ops = append(ops, op)
+		}
+	}
+
+	// Submit all the queued operations, using as many transactions as needed. (We're not really using
+	// transactions for atomicity, since we're not atomic anyway if there's more than one transaction,
+	// but batching them reduces the number of calls to the server.)
+	err = batchOperations(kv, ops)
+	return err
```
cwaldren-ld:
```suggestion
	return batchOperations(kv, ops)
```

sdktests/server_side_persistence_consul.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func batchOperations(kv *consul.KV, ops []*consul.KVTxnOp) error {
+	for i := 0; i < len(ops); {
```
cwaldren-ld, line 98: I know this comes from the consul integration, but it would be good if we had a comment explaining what it is actually doing at a high level.

sdktests/server_side_persistence_dynamodb.go
```diff
@@ -0,0 +1,246 @@
...
+	"strconv"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/launchdarkly/sdk-test-harness/v2/servicedef"
+)
+
+const (
+	// Schema of the DynamoDB table
+	tablePartitionKey = "namespace"
```
cwaldren-ld: Since these are all in the same package, you could access, say, `tableName` from the Redis file or whatever. It could be worth prefixing these to make them unique. 

Better yet would be refactoring these into dedicated packages, that way they can't start accidentally sharing constants. But that would be a fair amount of work. 

I'm surprised `initedKey` isn't being used in Redis impl (and is being used here)

sdktests/server_side_persistence_base.go
keelerm84: I like the optional string, so I'll go with that.

sdktests/server_side_persistence_consul.go
keelerm84: Well yes and no. I was wanting to show where it started and ended, so I was putting in some easy markers that you can jump between.

Are they also default vim markers? Yes, but that's beside the point. 😄 

Honestly, I was expecting more work in these files and figured it would be more useful than it is. So I will remove it.

sdktests/server_side_persistence_consul.go
keelerm84: I will update to be a pointer receiver.

sdktests/server_side_persistence_consul.go
keelerm84: I think I will leave this one to our refactoring work tomorrow.

sdktests/server_side_persistence_consul.go
keelerm84: Yes, I thought about this when I was copying this work over from the actual implementations. I went back and forth on whether consistency was better.

I'll just fix it here.

sdktests/server_side_persistence_dynamodb.go
keelerm84: Prefixed the dynamodb ones, and then moved the init one to the base file (I know that doesn't change anything, it's just an implication). Also changed its name and updated the usage in other places.

sdktests/server_side_persistence_base.go
```diff
@@ -6,13 +6,24 @@ import (
...
+
 	"github.com/launchdarkly/go-sdk-common/v3/ldcontext"
 	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
 	"github.com/launchdarkly/go-server-sdk-evaluation/v3/ldbuilders"
 	"github.com/launchdarkly/sdk-test-harness/v2/framework/ldtest"
 	"github.com/launchdarkly/sdk-test-harness/v2/servicedef"
 )
 
+const (
+	PersistenceInitedKey = "$inited"
```
cwaldren-ld: This is now a public constant accessible from other packages (capital `P`). Unclear if this is intended?

cwaldren-ld (approved):

cwaldren-ld (approved):

PR #249, opened 2024-10-24
chore: Add structure to automatically reset between test runs

keelerm84:


sdktests/server_side_persistence_base.go
```diff
@@ -1,6 +1,17 @@
 package sdktests
 
+// NOTE: You may note that these tests do not follow the same pattern as the
+// other tests in this repository.
+//
+// Historically, tests in this repository have suffered from onion-like
+// nesting. The further you have to dig, the more you cry.
+//
+// In the time honored tradition of every other language on the planet, we are
```
kinyoklion, line 9: image

sdktests/server_side_persistence_base.go
```diff
@@ -119,58 +134,560 @@ func newServerSidePersistentTests(
...
+			pollUntilFlagValueUpdated(t, client, "flag-key", context,
+				ldvalue.String("default"), ldvalue.String("fallthrough"), ldvalue.String("default"))
+
+			// Completely reset the database so there are no valid flag definitions
+			require.NoError(t, s.persistentStore.Reset())
+
+			h.RequireEventually(t,
+				checkForUpdatedValue(t, client, "flag-key", context,
+					ldvalue.String("fallthrough"), ldvalue.String("default"), ldvalue.String("default")),
+				time.Second, time.Millisecond*20, "flag value was NOT updated after cache TTL")
```
kinyoklion: I'm getting mixed messages between the assertion message and the test description.

sdktests/server_side_persistence_base.go
```diff
@@ -119,58 +134,560 @@ func newServerSidePersistentTests(
...
+		})
+
+		s.runWithEmptyStore(t, "caches flag forever", func(t *ldtest.T) {
+			persistence := NewPersistence()
+			persistence.SetStore(servicedef.SDKConfigPersistentStore{
+				Type: s.persistentStore.Type(),
+				DSN:  s.persistentStore.DSN(),
+			})
+			persistence.SetCache(servicedef.SDKConfigPersistentCache{
+				Mode: servicedef.CacheModeInfinite,
```
kinyoklion, line 301: Infinite mode isn't a uniform capability. I guess we can suppress the test though.

sdktests/server_side_persistence_base.go
keelerm84: Tweaked.

sdktests/server_side_persistence_base.go
keelerm84: That's a good point. I could make it a capability, but it does kind of feel like not supporting infinite is more of a suppression thing. If you can support the other two, seems like you should be able to support this.

sdktests/server_side_persistence_base.go
keelerm84: This was @cwaldren-ld 's prompting. Way to be the change you want to see. 😄

kinyoklion (approved):

PR #250, opened 2024-10-25
chore: Introduce opt-in flag to enable persistence support

keelerm84:
The SDK test harness has historically been self-sufficient. With the introduction of persistence-based unit tests, that is no longer the case. Externally started services are required for each persistence technology supported.

Since we want to retain the ease of use associated with a our current `go run .` usage, we are going to disable persistence support by default. Users will have to provide a `-enable-persistence-tests` if they want the test harness to attempt those tests.

SDKs under test are still required to provide the appropriate capabilities. In the event those capabilities are provided, but the test harness flag is not provided, persistence tests will be skipped.


docs/service_spec.md
```diff
@@ -102,11 +102,23 @@ v4 of the event schema originally required a `contextKeys` property on all featu
...
+- `consul`
+- `dynamodb`
+- `redis`
+
+The persistence store tests are unique in that they rely on external services to be available prior to running the test harness. Because of this, the test harness will not run these tests by default, even if the capabilities exist, without also provided the `-enable-persistence-tests` flag.
+
+These services can be easily provided through the use of docker. Example commands are shown below for your convenience.
+
+```shell
+docker run --network=host amazon/dynamodb-local
```
kinyoklion: For local tests you should be able to run these with network isolation? Not that most people aren't behind a NAT.

keelerm84: Right now the contract test is assuming everything is accessible on localhost:<default port>. So provided you meet that criteria, or we start adding more flexibility to the config options, then it doesn't matter how you set it up.

kinyoklion: I think `--network host` is experimental for mac and windows docker. Versus the normal port mapping.

```
The host networking driver only works on Linux hosts, but is available as a [beta feature](https://docs.docker.com/release-lifecycle/#beta) on Docker Desktop version 4.29 and later for Mac, Windows, and Linux. To enable this feature, navigate to the Resources tab in Settings, and then under Network select Enable host networking.
```

So, in theory:
`docker run -p 8000:8000 amazon/dynamodb-local`

Works on mac/windows.

keelerm84: Fixed.

kinyoklion (approved):

PR #253, opened 2024-10-29
fix: summary events should allow null default value in counter

cwaldren-ld:
When testing dotnet client, I came across the behavior that a `null` default value will end up in a summary event counter as the key `"default": null`.

This isn't accepted by the test harness, even though our JSON standard mandates that null and an omitted property be treated equivalently.

The SDK could fix this by omitting the property - but the harness should also accept this. 

-----

I'm open to hearing "no, the harness should not accept this."

### DIFF
--- a/sdktests/client_side_events_summary.go
+++ b/sdktests/client_side_events_summary.go
@@ -362,14 +362,14 @@ func doClientSideSummaryBasicPrereqTest(t *ldtest.T) {
 				m.KV("contextKinds", anyContextKindsList()),
 			)),
 			m.KV(prereq2Key, m.AllOf(
-				m.JSONOptProperty("default").Should(m.BeNil()),
+				JSONPropertyNullOrAbsent("default"),
 				m.JSONProperty("counters").Should(m.ItemsInAnyOrder(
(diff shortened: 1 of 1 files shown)


sdktests/client_side_events_summary.go
```diff
@@ -361,17 +361,19 @@ func doClientSideSummaryBasicPrereqTest(t *ldtest.T) {
 				)),
 				m.KV("contextKinds", anyContextKindsList()),
 			)),
-			m.KV(prereq2Key, m.MapOf(
-				m.KV("counters", m.ItemsInAnyOrder(
+			m.KV(prereq2Key, m.AllOf(
+				m.JSONOptProperty("default").Should(m.BeNil()),
```
kinyoklion: If this is fine, it makes me judge `JSONPropertyNullOrAbsent("value"),`

kinyoklion (approved):

sdktests/client_side_events_summary.go
kinyoklion: (It literally does the same thing, so I would vote for JSONPropertyNullOrAbsent).

cwaldren-ld: Oh, yup, had that before. Copy paste error.

kinyoklion (approved):

PR #254, opened 2024-10-31
feat: Introduce persistent store testing  support

keelerm84:


keelerm84:
Each of these individual commits have already been reviewed, except for the very last one (https://github.com/launchdarkly/sdk-test-harness/pull/254/commits/e755a822ddc170b5f1cdf882e04e101123ebdf47). That was something I just noticed doing final testing. Verified that change works though.

Do we want to squash this into a single commit, or just do some commit message overrides to get this released as a new feature?

cwaldren-ld (approved):
Seems like squash, unless you think we'll need to revert parts of it but not others.

PR #256, opened 2024-11-04
fix: Allow for more flexible flag key in tombstone

keelerm84:


sdktests/server_side_persistence_base.go
```diff
@@ -143,7 +143,7 @@ func newServerSidePersistentTests(
 }
 
 func (s *ServerSidePersistentTests) Run(t *ldtest.T) {
-	t.Run("uses default prefix", func(t *ldtest.T) {
+	s.runWithEmptyStore(t, "uses default prefix", func(t *ldtest.T) {
```
keelerm84, line 146: This was an issue I noticed when running the tests against a different local service, so I'm adding that on here.

sdktests/server_side_persistence_base.go
```diff
@@ -673,9 +673,12 @@ func basicFlagValidationMatcher(key string, version int, value string) m.Matcher
 	)
 }
 
-func basicDeletedFlagValidationMatcher(version int) m.Matcher {
+func basicDeletedFlagValidationMatcher(key string, version int) m.Matcher {
 	return m.AllOf(
-		m.JSONProperty("key").Should(m.Equal("$deleted")),
+		m.AnyOf(
+			m.JSONProperty("key").Should(m.Equal(key)),
```
keelerm84, line 679: I'm throwing this up here because it was quick to implement, but we should stop and decide if we actually want this.

Should we update each integration to require that the key, when deleted, is updated to $deleted?

kinyoklion (approved):

PR #258, opened 2024-11-13
feat: add streaming mode http proxy test

cwaldren-ld:
Adds a single test for streaming with HTTP proxy configuration (client-side). 

Ideally, the next bit of work will be..
- to support verifying that events and polling also go through the proxy.
- extending to server SDKs (should be just adding to the `server_side_stream_all.go`, but requires an SDK to test it)

The purpose of this test is to check that an SDK can make a request via a configured HTTP Proxy. 

In this mode, the SDK sends requests to the proxy, and the request body contains the intended URL that the proxy should connect to on behalf of the SDK. 

Future work might be to support other proxy modes, like HTTPS proxy (where the connection between SDK and proxy is encrypted) or SOCKS4/5.

### DIFF
--- a/docs/service_spec.md
+++ b/docs/service_spec.md
@@ -230,6 +230,14 @@ This means the SDK supports setting the wrapper name and version and includes th
 
 When this capability is set a `wrapper` configuration will be included with a subset of tests.
 
+
+### Capability `"http-proxy"`
+
+This indicates the SDK is capable of configuring an HTTP proxy for its network requests. 
+
+All requests should be sent to the proxy. This is generally implemented in an SDK via standard networking
+library capabilities, such as setting an environment variable (like `http_proxy`) or a configuration option.
+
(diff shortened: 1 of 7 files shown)


docs/service_spec.md
```diff
@@ -291,6 +299,8 @@ A `POST` request indicates that the test harness wants to start an instance of t
   * `wrapper` (object, optional): If specified contains wrapper configuration.
     * `name`: The name of the wrapper.
     * `version`: The version of the wrapper.
+  * `proxy` (object, optional): If specified contains proxy configuration.
+    * `httpProxy` (string, optional): An HTTP proxy, of the form `http://host:port`.
```
kinyoklion, line 303: I feel like this should be scheme, host, and port as independent fields. Easier to form a URL than parse one.

kinyoklion (approved):

PR #259, opened 2024-11-13
chore: Introduce DataSystem configuration

keelerm84:
This commit introduces the concept of the data system configuration. In addition, all existing polling and streaming configurations have been updated to be expressed in terms of primary synchronizers.

As it currently stands, most of the tests are passing. There are a handful that fail which I believe are expected. Even so, this is going into a feature branch, so I'm not too concerned about the failures just yet.


sdktests/client_side_stream_all.go
```diff
@@ -12,7 +12,6 @@ import (
 
 func doClientSideStreamTests(t *ldtest.T) {
 	t.Run("requests", doClientSideStreamRequestTest)
-	t.Run("updates", doClientSideStreamUpdateTests)
```
keelerm84, line 15: I removed these update tests from the client and server tests because they don't make sense in the fdv2 work. 

Originally these tests were making sure we handled individual flag / segment version values appropriately when updating the in memory store. That doesn't matter anymore (we strictly go off payload version), so we don't need this.

A subsequent commit will bring back something similar when we start interacting with the data stores. But they won't be written this like I would hope. 

--

This deletion goes until the next comment.

sdktests/server_side_stream_all.go
```diff
@@ -32,10 +31,6 @@ func doServerSideStreamRequestTests(t *ldtest.T) {
 	})
 }
 
-func doServerSideStreamUpdateTests(t *ldtest.T) {
-	NewCommonStreamingTests(t, "doServerSideStreamUpdateTests").Updates(t)
-}
-
```
keelerm84, line 38: end update cleanup.

sdktests/testapi_sdk_client.go
```diff
@@ -77,7 +77,14 @@ func WithEventsConfig(eventsConfig servicedef.SDKConfigEventParams) SDKConfigure
 // WithPollingConfig is used with StartSDKClient to specify a non-default polling configuration.
 func WithPollingConfig(pollingConfig servicedef.SDKConfigPollingParams) SDKConfigurer {
 	return helpers.ConfigOptionFunc[servicedef.SDKConfigParams](func(configOut *servicedef.SDKConfigParams) error {
-		configOut.Polling = o.Some(pollingConfig)
+		dataSystem := configOut.DataSystem.OrElse(servicedef.DataSystem{})
```
keelerm84, line 80: For FDv1 support, it is assumed that there is only ever polling or streaming. So we can safely assume, for this first iteration, that these must be equivalent to the primary synchronizer of FDV2 fame. 

Subsequent commits will rename these methods to be like `WithPrimaryPolling` or something of that nature. Letting that evolve as I refactor.

sdktests/testapi_sdk_data.go
```diff
@@ -148,15 +148,36 @@ func (d *SDKDataSource) Configure(config *servicedef.SDKConfigParams) error {
 	if d.streamingService == nil && d.pollingService == nil {
 		return errors.New("tried to use an SDKDataSource that has neither streaming nor polling configured")
 	}
+
 	if d.streamingService != nil {
-		newState := config.Streaming.Value()
-		newState.BaseURI = d.endpoint.BaseURL()
-		config.Streaming = o.Some(newState)
+		dataSystem := config.DataSystem.OrElse(servicedef.DataSystem{})
```
keelerm84, line 153: Ridiculous unpacking and re-assignment we discussed over Slack.

mockld/sdk_data.go
```diff
@@ -237,8 +237,8 @@ func (b *ServerSDKDataBuilder) BuildServerSDKData() ServerSDKData {
 	segments := maps.Clone(b.segments)
 
 	return map[DataItemKind]map[string]json.RawMessage{
-		"flags":    flags,
-		"segments": segments,
+		"flag":    flags,
```
cwaldren-ld, line 240: Why this change?

cwaldren-ld (approved):

mockld/sdk_data.go
keelerm84: In fdv1, that key was used as the data kind that was receiving the put object. It is still used that way in fdv2, but those keys are not pluralized.

PR #260, opened 2024-11-14
chore: Remove top level streaming, polling, and service endpoint configurations

keelerm84:
With the new FDv2 setup, we no longer have top level streaming and polling configurations. Everything is configured through the new data system configuration.

Because of this change, supporting service endpoints as a concept no longer really applies. It is totally reasonable to think you might want to initialize a poller from a different URL than the synchronizer. So having a single top level configuration for those endpoints doesn't align with the future structure. 

So we are removing all of that.


sdktests/testapi_sdk_client.go
```diff
@@ -204,25 +195,13 @@ func TryNewSDKClient(t *ldtest.T, configurers ...SDKConfigurer) (*SDKClient, err
...
-	// TODO: Add some validation here for the FDv2 config
-	if !config.Streaming.IsDefined() && !config.Polling.IsDefined() && !config.DataSystem.IsDefined() &&
-		!config.PersistentDataStore.IsDefined() && config.ServiceEndpoints.Value().Streaming == "" {
+	if !config.DataSystem.IsDefined() && !config.PersistentDataStore.IsDefined() {
 		// Note that the default is streaming, so we don't necessarily need to set config.Streaming if there are
 		// no other customized options and if we used serviceEndpoints.streaming to set the stream URI
 		return errors.New(
 			"neither streaming nor polling was enabled-- did you forget to include the SDKDataSource as a parameter?")
 	}
-	if config.Streaming.IsDefined() && config.Streaming.Value().BaseURI == "" &&
```
cwaldren-ld, line 216: Well this deletion feels good. Very good.

cwaldren-ld (approved):

PR #264, opened 2024-11-18
feat: add http proxy test for client-side polling

cwaldren-ld:


sdktests/client_side_stream_all.go
```diff
@@ -22,7 +23,9 @@ func doClientSideStreamRequestTest(t *ldtest.T) {
 	streamTests := NewCommonStreamingTests(t, "doClientSideStreamRequestTest",
 		WithCredential(envIDOrMobileKey))
 
-	streamTests.RequestViaHTTPProxy(t)
```
cwaldren-ld, line 25: Just hoisting the capability check upwards, as is the pattern. Well, a pattern. Other tests put the check directly in the test 🤷‍♂️

kinyoklion (approved):

PR #268, opened 2024-11-22
chore: Add tests for polling initialization

keelerm84:


keelerm84:
@cwaldren-ld for these tests to pass, we also need the modifications in https://github.com/launchdarkly/go-server-sdk/pull/205

cwaldren-ld (approved):

PR #274, opened 2025-01-03
fix: Remove invalid conditions from persistent tests

keelerm84:


sdktests/server_side_persistence_base.go
```diff
@@ -572,72 +572,10 @@ func (s *ServerSidePersistentTests) Run(t *ldtest.T) {
 
 					require.NoError(t, s.persistentStore.WriteMap(s.defaultPrefix, "features", s.initialFlags))
 
-					if cacheConfig.Mode == servicedef.CacheModeInfinite {
-						// This key was already cached, so it shouldn't see the change above.
-						h.RequireNever(t,
-							checkForUpdatedValue(t, client, "flag-key", context,
-								ldvalue.String("value"), ldvalue.String("new-value"), ldvalue.String("default")),
-							time.Millisecond*500, time.Millisecond*20, "flag-key was incorrectly updated")
```
keelerm84, line 580: The diff looks weird, but basically the first two if / else if blocks are reduced to this single check.

In FDv2, if the data system has a data source that has provided data, we never rely on the  persistent store again. As a result, the "But since ..." checks weren't valid.

sdktests/server_side_persistence_base.go
```diff
@@ -572,72 +572,10 @@ func (s *ServerSidePersistentTests) Run(t *ldtest.T) {
...
-
-						// But eventually, it will expire and then we will fetch it from the database.
-						h.RequireEventually(t,
-							checkForUpdatedValue(t, client, "flag-key", context,
-								ldvalue.String("value"), ldvalue.String("fallthrough"), ldvalue.String("default")),
-							time.Duration(int(time.Second)*cacheConfig.TTL.Value()), time.Millisecond*20, "flag-key was incorrectly cached")
-					}
-				})
-
-				s.runWithEmptyStore(t, "ignores dropped flags", func(t *ldtest.T) {
```
keelerm84, line 606: This is redundant with the check above. Whether I delete flags or change them, it is confirming the same thing.

tanderson-ld (approved):

PR #294, opened 2025-04-16
feat: Add support for client-side per-context summary events.

kinyoklion:


docs/service_spec.md
```diff
@@ -248,6 +248,10 @@ This indicates the SDK is capable of configuring an HTTP proxy for its network r
 All requests should be sent to the proxy. This is generally implemented in an SDK via standard networking
 library capabilities, such as setting an environment variable (like `http_proxy`) or a configuration option.
 
+### Capability `"client-per-context-summaries`
+
+This indicates that a client-side SDK will emit a summary event per identified context which has evaluated flags.
```
tanderson-ld, line 253: Would you like to include a link to the spec?

tanderson-ld (approved):

docs/service_spec.md
kinyoklion: Hmm. This repo is public and most of the specs are not, there is a public version of this spec, but the public one isn't synced yet. So, at the moment, no. I like the idea though.

PR #304, opened 2025-06-11
chore: adds ConvertToFDv2SDKClientData to handle client flags to FDv2…

tanderson-ld:


mockld/sdk_data.go
```diff
@@ -113,6 +113,31 @@ func (d ServerSDKData) ConvertToFDv2SDKData(t *ldtest.T) FDv2SDKData {
...
+		payloadObjects = append(payloadObjects, framework.BaseObject{
+			Kind:    "flag",
+			Version: 1,
+			Key:     key,
+			Object:  json,
+		})
+	}
+
+	return FDv2SDKData{
+		intentCode:   "xfer-full",
```
keelerm84, line 134: The sdk has a subsystems.IntentTransferFull constant you can use.

mockld/sdk_data.go
```diff
@@ -113,6 +113,31 @@ func (d ServerSDKData) ConvertToFDv2SDKData(t *ldtest.T) FDv2SDKData {
...
+	payloadObjects := make([]framework.BaseObject, 0)
+
+	for key, item := range d {
+		json, err := json.Marshal(item)
+		if err != nil {
+			panic(err)
+		}
+
+		payloadObjects = append(payloadObjects, framework.BaseObject{
+			Kind:    "flag",
```
keelerm84, line 126: There is a FlagKind constant you can use as well.

mockld/sdk_data.go
```diff
@@ -113,6 +113,31 @@ func (d ServerSDKData) ConvertToFDv2SDKData(t *ldtest.T) FDv2SDKData {
...
+			Version: 1,
```
keelerm84: Should this be the actual version of the item?

mockld/sdk_data.go
```diff
@@ -113,6 +113,31 @@ func (d ServerSDKData) ConvertToFDv2SDKData(t *ldtest.T) FDv2SDKData {
 // This does not include flag or segment configurations, but only flag evaluation results for a specific user.
 type ClientSDKData map[string]ClientSDKFlag
 
+func (d ClientSDKData) ConvertToFDv2SDKClientData(t *ldtest.T) FDv2SDKData {
```
keelerm84: I'm doing some tweaks to go and have a similar method. I found it helpful to accept a selector here. Then you can control the final event output.

mockld/sdk_data.go
tanderson-ld: Done

mockld/sdk_data.go
tanderson-ld: Done

mockld/sdk_data.go
tanderson-ld: Good catch, fixed.

mockld/sdk_data.go
tanderson-ld: Discussed on slack, not doing to avoid additional go-server sdk dependency.

mockld/sdk_data.go
tanderson-ld: Discussed on slack, not doing to avoid additional go-server sdk dependency.

mockld/sdk_data.go
tanderson-ld: Done

PR #311, opened 2026-01-26
fix: Init the persistence store prior to client creation

jsonbailey:
Tracking internally as SDK-1784

### DIFF
--- a/sdktests/server_side_persistence_base.go
+++ b/sdktests/server_side_persistence_base.go
@@ -381,13 +381,13 @@ func (s *ServerSidePersistentTests) Run(t *ldtest.T) {
 					closeWhenReady := make(chan bool)
 					endpoint := blockingEndpoint(closeWhenReady, dataSystem.PrimarySync().streaming)
 
+					require.NoError(t, s.persistentStore.WriteMap(s.defaultPrefix, "features", s.initialFlags))
+					require.NoError(t, s.persistentStore.Write(s.defaultPrefix, persistenceInitedKey, "1"))
+
 					client := NewSDKClient(t,
 						persistence,
 						WithWaitToStart(time.Millisecond, true),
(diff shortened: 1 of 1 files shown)


sdktests/server_side_persistence_base.go
```diff
@@ -381,14 +381,14 @@ func (s *ServerSidePersistentTests) Run(t *ldtest.T) {
 					closeWhenReady := make(chan bool)
 					endpoint := blockingEndpoint(closeWhenReady, dataSystem.PrimarySync().streaming)
 
+					require.NoError(t, s.persistentStore.WriteMap(s.defaultPrefix, "features", s.initialFlags))
```
jsonbailey, line 384: I can't get this test to pass in Ruby without this change and I believe it is a test issue / race condition. When the Ruby SDK initializes it checks if the persistence store is inited on startup and that value will be cached based on the options passed which for this test are No Cache, 1 second Cache, and Forever cache.

This means the first check when we are using cache will return the default value and not the value from the persistence store which is what this test expects causing a failure. If instead we ensure the persistence store is setup, it should initialize with a valid persistence store regardless of cache setting, and show that we properly switch to the data source after we unblock streaming.

keelerm84: Yes, this seems like a reasonable change to me.

Did you check the other tests in this file? If I did it out of order here, I'm sure it's like that in the other tests.

keelerm84 (approved):

sdktests/server_side_persistence_base.go
jsonbailey: I did look at the other tests and they all seem fine. When they write to redis after client creation, they often have cache mode disabled, which prevents the same error, or cache mode to 1 second and they are explicitly testing that after a second they will pull in the data written after client creation.

PR #313, opened 2026-01-29
fix: Send synchonizers as a list

jsonbailey:


sdktests/testapi_sdk_data_system.go
```diff
@@ -47,29 +47,19 @@ func DataSystemOptionStreaming() SDKDataSystemOption {
...
-		s.Synchronizers = &Synchronizers{}
-	}
-	s.Synchronizers.secondary = &synchronizer
+func (s *SDKDataSystem) AddDataSynchronizer(synchronizer DataSynchronizer) {
+	s.Synchronizers = append(s.Synchronizers, synchronizer)
 }
 
-func (s *SDKDataSystem) PrimarySync() *Synchronizer {
-	return &s.Synchronizers.primary
+func (s *SDKDataSystem) PrimarySync() *DataSynchronizer {
```
jsonbailey: For initializers we just reference them directly `dataSystem.Initializers[0].Endpoint().AwaitConnection(time.Second)` vs `dataSystem.PrimarySync().Endpoint().RequireConnection(t, time.Second)`

This PrimarySync method feels unnecessary with this change but is used a lot. Its an easy update but I wanted a second opinion before making that change.

keelerm84: We only even started with those methods because they match the way we were talking about it internally. But that isn't the case anymore, so I would be fine with going to direct indexing.

keelerm84 (approved):

PR #315, opened 2026-02-10
feat: adds long running tests for polling intervals

tanderson-ld:
This PR adds 3 tests for testing polling intervals.  They are long running tests and so can only be run while passing the -enable-long-running-tests flag.

### DIFF
--- a/sdktests/common_tests_poll_interval.go
+++ b/sdktests/common_tests_poll_interval.go
@@ -28,7 +28,8 @@ func (c CommonPollingTests) pollIntervals(t *ldtest.T) struct{ Default, Min time
 	panic("pollIntervals: test service has neither server-side nor client-side capability")
 }
 
-// Requires -enable-long-running-tests. Only runs when client-side or server-side-polling capability is present.
+// PollingIntervalTests requires -enable-long-running-tests and only runs when client-side or
+// server-side-polling capability is present.
 func (c CommonPollingTests) PollingIntervalTests(t *ldtest.T) {
 	if t.Capabilities().HasAny(servicedef.CapabilityClientSide, servicedef.CapabilityServerSidePolling) {
 		t.LongRunning()
--- a/sdktests/common_tests_poll_interval.go
+++ b/sdktests/common_tests_poll_interval.go
@@ -7,6 +7,7 @@ import (
 	"github.com/launchdarkly/sdk-test-harness/v2/framework/ldtest"
 	"github.com/launchdarkly/sdk-test-harness/v2/mockld"
 	"github.com/launchdarkly/sdk-test-harness/v2/servicedef"
+	"github.com/stretchr/testify/require"
 )
 
 const pollIntervalEpsilon = 2 * time.Second
@@ -26,6 +27,7 @@ func (c CommonPollingTests) pollIntervals(t *ldtest.T) struct{ Default, Min time
 		return struct{ Default, Min time.Duration }{Default: 5 * time.Minute, Min: 5 * time.Minute}
 	}
 	require.Fail(t, "pollIntervals: test service has neither server-side nor client-side capability")
+	return struct{ Default, Min time.Duration }{} // unreachable; satisfy compiler
 }
 
 // PollingIntervalTests requires -enable-long-running-tests and only runs when client-side or
--- a/sdktests/common_tests_poll_interval.go
+++ b/sdktests/common_tests_poll_interval.go
@@ -1,0 +1,120 @@
+package sdktests
+
+import (
+	"time"
+
+	"github.com/launchdarkly/sdk-test-harness/v2/framework/harness"
+	"github.com/launchdarkly/sdk-test-harness/v2/framework/ldtest"
+	"github.com/launchdarkly/sdk-test-harness/v2/mockld"
+	"github.com/launchdarkly/sdk-test-harness/v2/servicedef"
+)
+
+const pollIntervalEpsilon = 2 * time.Second
+
+// pollIntervals returns the interval parameters for the polling interval tests (Default and Min). Uses server-side/client-side capabilities; uses sdkKind only for the browser (JSClientSDK) special case. Panics if no condition matches. Return type is an anonymous struct so it is not referenced from other files.
+func (c CommonPollingTests) pollIntervals(t *ldtest.T) struct{ Default, Min time.Duration } {
+	if t.Capabilities().Has(servicedef.CapabilityServerSide) {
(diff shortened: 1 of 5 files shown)


sdktests/common_tests_poll_interval.go
```diff
@@ -0,0 +1,125 @@
...
+		return struct{ Default, Min time.Duration }{Default: 30 * time.Second, Min: 30 * time.Second}
+	}
+	// Browser (JS) SDKs use 5min default, 30s min.
+	if c.sdkKind == mockld.JSClientSDK {
+		return struct{ Default, Min time.Duration }{Default: 5 * time.Minute, Min: 30 * time.Second}
+	}
+	if t.Capabilities().Has(servicedef.CapabilityClientSide) {
+		return struct{ Default, Min time.Duration }{Default: 5 * time.Minute, Min: 5 * time.Minute}
+	}
+	panic("pollIntervals: test service has neither server-side nor client-side capability")
```
aaron-zeisler: Question: Is there a `t.Fail()` command that could be used here instead of `panic()`? I'm concerned that the panic() may completely stop processing, preventing any other tests from executing. I'm not sure if that's true for this framework, I'd have to see it fail in order to be sure, but generally I think a t.Fail() (or returning an error) is better than a panic() unless we absolutely want all the code to stop executing.

sdktests/common_tests_poll_interval.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/launchdarkly/sdk-test-harness/v2/mockld"
+	"github.com/launchdarkly/sdk-test-harness/v2/servicedef"
+)
+
+const pollIntervalEpsilon = 2 * time.Second
+
+// pollIntervals returns the interval parameters for the polling interval tests (Default and Min).
+// Uses server-side/client-side capabilities; uses sdkKind only for the browser (JSClientSDK) special case.
+// Panics if no condition matches. Return type is an anonymous struct so it is not referenced from other files.
+func (c CommonPollingTests) pollIntervals(t *ldtest.T) struct{ Default, Min time.Duration } {
```
aaron-zeisler, line 19: Mega-nit-picky: I haven't seen enough code to understand if this is a common convention, but in my opinion, defining a struct type for this function's return value would make the code more readable. The struct could be non-exported (ie it could start with a lower-case letter) so only this `sdktests` package can use it.

sdktests/common_tests_poll_interval.go
```diff
@@ -0,0 +1,125 @@
...
+	if t.Capabilities().Has(servicedef.CapabilityServerSide) {
+		return struct{ Default, Min time.Duration }{Default: 30 * time.Second, Min: 30 * time.Second}
+	}
+	// Browser (JS) SDKs use 5min default, 30s min.
+	if c.sdkKind == mockld.JSClientSDK {
+		return struct{ Default, Min time.Duration }{Default: 5 * time.Minute, Min: 30 * time.Second}
+	}
+	if t.Capabilities().Has(servicedef.CapabilityClientSide) {
+		return struct{ Default, Min time.Duration }{Default: 5 * time.Minute, Min: 5 * time.Minute}
+	}
```
aaron-zeisler, line 29: Curiosity questions: I'm sure this has been discussed :) But is it possible to inject smaller polling configurations into the SDK that's under test? For example, could we inject a polling interval of 10 seconds (or less) to speed up these tests while still verifying they behave as we expect?

Alternatively, is it possible to retrieve the polling configuration values from the SDK under test, rather than hard-coding them in this function?

aaron-zeisler: Ah, I see, it looks like we can set the SDK's polling configuration using the `DataSystemOptionPollInterval` option in NewSDKDataSystem(). Still curious about the pros/cons of using the actual polling intervals rather than mocked values.

aaron-zeisler (approved):
The logic looks good to me. I have a bunch of questions and optional/minor suggestions but no blockers.

When we pair up, can you please show me how to run these tests? Apologies if you've showed me and I forgot already.

sdktests/common_tests_poll_interval.go
tanderson-ld: The SDKs have logic in them to clamp the minimum polling interval.  If the test harness could set the polling interval to less than the minimum for testing, then it is probably the case a customer could as well unless we create some scheme to give the sdk test harness elevated privileges.

sdktests/common_tests_poll_interval.go
tanderson-ld: I initially had it that way, but I was very worried another piece of code could be added at some point that depends on the struct.  I didn't want to create a harness wide norm around describing these defaults yet.

sdktests/common_tests_poll_interval.go
tanderson-ld: I wanted this to appear as a developer error.  Getting here indicates the tests weren't ready to handle this SDK because it doesn't know what such an SDKs default should be.

sdktests/common_tests_poll_interval.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"github.com/launchdarkly/sdk-test-harness/v2/framework/harness"
+	"github.com/launchdarkly/sdk-test-harness/v2/framework/ldtest"
+	"github.com/launchdarkly/sdk-test-harness/v2/mockld"
+	"github.com/launchdarkly/sdk-test-harness/v2/servicedef"
+)
+
+const pollIntervalEpsilon = 2 * time.Second
+
+// pollIntervals returns the interval parameters for the polling interval tests (Default and Min).
```
keelerm84: It might be me, but I felt that first sentence read oddly.

```suggestion
// pollIntervals returns the default and minimum values used for interval testing.
```

sdktests/common_tests_poll_interval.go
keelerm84: Yes, I believe it's something like:

```suggestion
	require.Fail(t, "pollIntervals: test service has neither server-side nor client-side capability")
```

sdktests/common_tests_poll_interval.go
keelerm84: We do allow you to define polling intervals, but the SDKs enforce a 30 second minimum polling speed.

If we allow the test harness to override that and set a lower value, then we effectively allow customers to do the same. This could lead to someone slamming out service without intentionally doing so. And since that affects their billing, it could be costly for both parties.

It's true customers could still write their own data source and do this anyway. There are other remediation in place for that too. But we don't want to make it easy for a good customer to make a costly choice.

keelerm84 (approved):

PR #336, opened 2026-04-24
feat: Add FDv1 Fallback Directive tests and config

keelerm84:
Adds a new "FDv1 fallback directive" test suite covering section 1.6 of
the Data System spec: directive on streaming error, directive on
streaming success with payload-first handling, directive on a polling
initializer skipping the FDv2 synchronizer chain, halt behavior when no
FDv1 fallback is configured, and one-way terminality of Directed Fallback.

Introduces FDv1Fallback as a first-class field on DataSystem, with
dedicated WithFDv1FallbackPollingSynchronizer and
WithFDv1FallbackStreamingSynchronizer configurers. Previously the only
way to signal an FDv1 fallback URL was to overload the last entry of
Synchronizers, which misrepresented the SDK's architecture — the FDv1
Fallback Synchronizer is distinct from the FDv2 Primary/Fallback chain
and must be configured separately.

Tests assert evaluation provenance, not just URL hits: the "winning"
data source serves a distinctive flag value and the "losing" side
responds with 304 or an unrelated error, so a successful evaluation of
the expected value proves the data actually flowed through the intended
path. Gated on a new "fdv1-fallback" capability so SDKs that have not
yet implemented the directive opt out and the suite can be
decommissioned by dropping the capability once ubiquitous.

---

> [!NOTE]
> **Medium Risk**
> Adds a new directed-fallback test suite and extends the shared `DataSystem` config schema with `fdv1Fallback`, which may require updates in SDK test services and could break consumers expecting the old JSON shape.
> 
> **Overview**
> Adds a new *FDv1 Fallback Directive* test suite (Data System spec §1.6) that validates switching to an FDv1 polling endpoint when `X-LD-FD-Fallback: true` is returned, including payload-before-switch behavior, skipping FDv2 synchronizers when directed by an initializer, halting when no FDv1 fallback is configured, and ensuring the switch is terminal (no recovery back to FDv2).
> 
> Introduces first-class configuration for FDv1 fallback by adding `DataSystem.fdv1Fallback` and a `WithFDv1Fallback` helper, plus a new `fdv1-fallback` capability gate to allow SDKs to opt in to these tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9b8a181ff83b9d65df86fe6a214c46f46e467d51. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sdktests/common_tests_stream_fdv2.go
+++ b/sdktests/common_tests_stream_fdv2.go
@@ -450,6 +450,11 @@ func (c CommonStreamingTests) DirectiveOnStreamingErrorEngagesFDv1(t *ldtest.T) 
 			BaseURI: fdv1Endpoint.BaseURL(),
 		}))
 
+	// Drain the initial FDv2 connection so the quiescence check below has a
+	// stable baseline.
+	_, err := streamEndpoint.AwaitConnection(time.Second)
+	require.NoError(t, err)
+
 	// FDv1 endpoint must be contacted at /sdk/latest-all.
 	h.RequireEventually(t, func() bool {
 		select {
@@ -469,6 +474,15 @@ func (c CommonStreamingTests) DirectiveOnStreamingErrorEngagesFDv1(t *ldtest.T) 
 		return m.In(t).Assert(value, m.JSONEqual(fdv1Value))
 	}, time.Second*3, time.Millisecond*20,
 		"flag-key should have been served from the FDv1 fallback after directive on streaming error")
+
+	// FDv2 streaming must be quiet — no concurrent retries against the (now
+	// halted) Primary Synchronizer. Per 1.6.3(2), the FDv2 chain must be stopped
+	// once the directive engages the FDv1 fallback, so observing a second
+	// connection here would mean the SDK is running both data sources in
+	// parallel. 403 alone would already have caused permanent removal, so this
+	// check is defensive — it catches the case where the SDK permanently removes
+	// one synchronizer but still has a background loop dialing the endpoint.
+	streamEndpoint.RequireNoMoreConnections(t, time.Millisecond*500)
 }
 
 // DirectiveOnStreamingSuccessAppliesPayload verifies Requirement 1.6.2: when the
--- a/sdktests/common_tests_stream_fdv2.go
+++ b/sdktests/common_tests_stream_fdv2.go
@@ -1,6 +1,8 @@
 package sdktests
 
 import (
+	"encoding/json"
+	"fmt"
 	"net/http"
 	"time"
 
@@ -19,6 +21,25 @@ import (
 
 	"github.com/stretchr/testify/require"
 )
+
+// serverSideFDv1AllData builds an FDv1-format polling response body containing a single flag.
+// The FDv1 wire format is a single JSON object with `"flags"` and `"segments"` keys, distinct
+// from the FDv2 polling wire format. Tests use this to stand up a realistic FDv1 Fallback
+// Synchronizer endpoint.
+func serverSideFDv1AllData(t *ldtest.T, c CommonStreamingTests, flagKey string, version int, value ldvalue.Value) []byte {
+	t.Helper()
+	body := map[string]any{
+		"flags": map[string]json.RawMessage{
+			flagKey: c.makeFlagData(flagKey, version, value),
+		},
+		"segments": map[string]json.RawMessage{},
+	}
+	bytes, err := json.Marshal(body)
+	if err != nil {
+		panic(fmt.Errorf("failed to marshal FDv1 polling body: %w", err))
+	}
+	return bytes
+}
 
 var (
 	initialValue = ldvalue.String("initial value") //nolint:gochecknoglobals
@@ -46,7 +67,7 @@ func (c CommonStreamingTests) FDv2(t *ldtest.T) {
 	t.Run("permanent fallback to secondary synchronizer", c.PermanentFallbackToSecondarySynchronizer)
 	t.Run("recoverable fallback with recovery", c.RecoverableFallbackWithRecovery)
 	t.Run("permanent fallback with recovery", c.PermanentFallbackWithRecovery)
-	t.Run("fallback to FDv1 handling", c.FallbackFromFDv2ToFDv1)
+	t.Run("FDv1 fallback directive", c.FDv1FallbackDirective)
 }
 
 func (c CommonStreamingTests) StateTransitions(t *ldtest.T) {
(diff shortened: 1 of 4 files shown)


sdktests/common_tests_stream_fdv2.go
```diff
@@ -46,7 +67,7 @@ func (c CommonStreamingTests) FDv2(t *ldtest.T) {
 	t.Run("permanent fallback to secondary synchronizer", c.PermanentFallbackToSecondarySynchronizer)
 	t.Run("recoverable fallback with recovery", c.RecoverableFallbackWithRecovery)
 	t.Run("permanent fallback with recovery", c.PermanentFallbackWithRecovery)
-	t.Run("fallback to FDv1 handling", c.FallbackFromFDv2ToFDv1)
```
keelerm84, line 49: Expanding from a single test to more detailed test cases.

sdktests/common_tests_stream_fdv2.go
```diff
@@ -366,38 +387,400 @@ func (c CommonStreamingTests) PermanentFallbackWithRecovery(t *ldtest.T) {
...
+
+// FDv1FallbackDirective groups the server-directed FDv1 fallback tests defined by
+// section 1.6 of the Data System spec. See:
+// specs/DATASYSTEM-data-system/v2/README.md#16-fdv1-fallback-directive
+//
+// These tests are gated on the "fdv1-fallback" capability so SDKs that have not yet
+// implemented the directive can opt out. Once the behavior is ubiquitous across SDKs,
+// dropping the capability from every test service decommissions the suite cleanly.
+func (c CommonStreamingTests) FDv1FallbackDirective(t *ldtest.T) {
+	t.RequireCapability(servicedef.CapabilityFDv1Fallback)
```
keelerm84, line 408: All of these tests are gated by a new capability. This is because we are going to change the shape of the config required for exercising these tests and this ensures we don't break existing SDKs until they can add this capability.

sdktests/common_tests_stream_fdv2.go
```diff
@@ -366,38 +387,400 @@ func (c CommonStreamingTests) PermanentFallbackWithRecovery(t *ldtest.T) {
...
+		c.DirectiveOnStreamingSuccessAppliesPayload)
+	t.Run("directive on polling initializer skips FDv2 synchronizers",
+		c.DirectiveOnPollingInitializerSkipsSynchronizers)
+	t.Run("directive without FDv1 fallback configured halts the data system",
+		c.DirectiveWithoutFDv1ConfiguredHaltsDataSystem)
+	t.Run("directed fallback is terminal and does not revisit FDv2 sources",
+		c.DirectedFallbackIsTerminal)
+}
+
+// DirectiveOnStreamingErrorEngagesFDv1 verifies Requirement 1.6.1 and 1.6.3 for the
```
keelerm84, line 422: These spec requirements relate to the [spec changes here](https://github.com/launchdarkly/sdk-specs/pull/155), which cover the fallback behavior more specifically.

servicedef/sdk_config.go
```diff
@@ -39,11 +39,12 @@ const (
...
-	Store         o.Maybe[DataStore] `json:"store,omitempty"`
-	StoreMode     DataStoreMode      `json:"storeMode"`
-	Initializers  []DataInitializer  `json:"initializers"`
-	Synchronizers []DataSynchronizer `json:"synchronizers"`
-	PayloadFilter o.Maybe[string]    `json:"payloadFilter,omitempty"`
+	Store         o.Maybe[DataStore]                `json:"store,omitempty"`
+	StoreMode     DataStoreMode                     `json:"storeMode"`
+	Initializers  []DataInitializer                 `json:"initializers"`
+	Synchronizers []DataSynchronizer                `json:"synchronizers"`
+	FDv1Fallback  o.Maybe[FDv1FallbackSynchronizer] `json:"fdv1Fallback,omitempty"`
```
keelerm84: This is the new top level config change required to support the fdv1 capability. Before, the SDKs were having to guess at what the connection URL should be, or how exactly it should be configured. This allows for much more explicit control.

sdktests/common_tests_stream_fdv2.go
```diff
@@ -366,38 +387,398 @@ func (c CommonStreamingTests) PermanentFallbackWithRecovery(t *ldtest.T) {
...
+		select {
+		case resp := <-fdv1Channel:
+			return resp.Request.URL.Path == fdv1PollingPath
+		default:
+			return false
+		}
+	}, time.Second*3, time.Millisecond*10,
+		"FDv1 fallback endpoint was never contacted after directive on streaming error")
+
+	// Evaluation must reflect FDv1's payload. If we saw the default here, the SDK
```
kinyoklion, line 473: Do we need an assertion to require no connection on the FDv2 endpoint? To make sure it isn't concurrently retrying FDv2? (Like you didn't close the data source).

sdktests/common_tests_stream_fdv2.go
```diff
@@ -366,38 +387,398 @@ func (c CommonStreamingTests) PermanentFallbackWithRecovery(t *ldtest.T) {
...
+		"flag-key should have been served from the FDv1 fallback after directive on streaming error")
+}
+
+// DirectiveOnStreamingSuccessAppliesPayload verifies Requirement 1.6.2: when the
+// directive arrives alongside a valid payload (here, a 200 SSE handshake that delivers
+// a full basis), the SDK MUST apply that payload to the Memory Store before surfacing
+// the directive and transitioning to the FDv1 Fallback Synchronizer.
+//
+// The streaming endpoint serves a distinctive value (streamingValue). The FDv1
+// fallback endpoint responds with 304 Not Modified for every request, i.e. "your
```
kinyoklion: This is a semantic violation. I am not actually sure what would happen if you expose a browser to this.

Like we have to have sent, for that specific request/if-none-match a 2xx code previously. As the browser would handle a 304 and then use its cached value to translate that into a 2xx.

tanderson-ld: Is a 200 with transfer none more clear and gets around the browser behavior?

tanderson-ld (approved):
kinyoklion has an open comment when I submitted this approval.

sdktests/common_tests_stream_fdv2.go
keelerm84: This is the fdv1 endpoint. There is no `transfer-none` on that protocol. I'll update it to return a 400. That accomplishes the same thing.

sdktests/common_tests_stream_fdv2.go
keelerm84: Updated to check this.

kinyoklion (approved):

PR #340, opened 2026-05-15
test: Cover FDv2 request shapes in instance-id and tags tests

keelerm84:
## Summary

`CommonInstanceIDTests` and `CommonTagsTests` previously verified the `X-LaunchDarkly-Instance-Id` and `X-LaunchDarkly-Tags` headers only on a single streaming synchronizer, a single polling synchronizer, and the events endpoint. FDv2 introduced three new request shapes that the suite did not exercise:

- Polling Initializers that run before the synchronizer (`DataSystemOptionPollingInitializer`).
- Secondary Synchronizers reached after the Primary is permanently removed (the FDv2 fallback chain in `CommonStreamingTests.FDv2`).
- The FDv1 Fallback Synchronizer reached via the server-directed FDv1 Fallback Directive (#336).

A regression that dropped either header from any of these new request shapes would not have been caught by the existing tests.

This PR adds three subtests to each of the two files, standing up each request shape and asserting the relevant header is present:

- `polling initializer requests` -- builds a data system with a polling initializer plus a no-op (`xfer-none` / `up-to-date`) synchronizer; asserts on the initializer endpoint.
- `secondary synchronizer requests after permanent fallback` -- primary returns 401 (non-recoverable, immediate permanent removal); secondary serves an empty FDv2 stream; asserts on the secondary endpoint.
- `FDv1 fallback directive requests` -- FDv2 streaming returns `403 + X-LD-FD-Fallback: true`; FDv1 endpoint serves an empty `{"flags":{},"segments":{}}` body so init completes along the fallback path; asserts on the FDv1 endpoint.

The new subtests are server-side only -- the FDv2 features they exercise are server-side today, matching the gating used in `server_side_stream_all.go`. The FDv1 directive subtest is additionally gated on `CapabilityFDv1Fallback` so SDKs that have not yet implemented the directive can opt out.

The tags variant uses one representative tags config for the new subtests rather than iterating over `makeValidTagsTestParams` -- endpoint coverage and tag-value variation are orthogonal properties, and the existing subtests already cover the latter.

---

> [!NOTE]
> **Low Risk**
> Test-only changes to the SDK test harness; no production SDK or service behavior is modified.
> 
> **Overview**
> Expands the SDK test harness so **`X-LaunchDarkly-Instance-Id`** and **`X-LaunchDarkly-Tags`** are verified on FDv2-related request paths that were previously untested (streaming/poll/events only).
> 
> **Instance ID:** Replaces one-off header checks with **`newInstanceIDChecker`**, which requires a non-empty header and the **same** value across every endpoint touched by one client. Adds a subtest that **distinct clients get distinct** instance IDs (skipped when `CapabilitySingleton`). Adds server-side subtests for **polling initializer**, **secondary synchronizer after permanent 401 fallback**, and **FDv1 fallback directive** (gated on `CapabilityFDv1Fallback`), including assertions that abandoned endpoints stop receiving traffic.
> 
> **Tags:** Mirrors the same three FDv2 request-shape subtests using one representative tags config (endpoint coverage vs tag permutations).
> 
> **Shared:** Adds **`emptyFDv1FallbackBody`** on `commonTestsBase` so FDv1 fallback mocks return server- vs client-side-shaped empty JSON for init-only scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 48e0fd903b1cb3eb6263bdb048ba237e75c7b11b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/sdktests/common_tests_instance_id.go
+++ b/sdktests/common_tests_instance_id.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	"github.com/launchdarkly/go-sdk-common/v3/ldcontext"
-	"github.com/launchdarkly/go-sdk-common/v3/ldtime"
 	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
 	"github.com/launchdarkly/go-test-helpers/v2/httphelpers"
 	"github.com/launchdarkly/sdk-test-harness/v2/framework/harness"
@@ -174,9 +173,7 @@ func (c CommonInstanceIDTests) Run(t *ldtest.T) {
 			t.Defer(fdv1Endpoint.Close)
 
 			_ = NewSDKClient(t, c.baseSDKConfigurationPlus(
-				WithConfig(servicedef.SDKConfigParams{
-					StartWaitTimeMS: o.Some(ldtime.UnixMillisecondTime(5000)),
-				}),
+				WithWaitToStart(5*time.Second, false),
 				WithStreamingSynchronizer(servicedef.SDKConfigStreamingParams{
 					BaseURI: streamEndpoint.BaseURL(),
 				}),
--- a/sdktests/common_tests_instance_id.go
+++ b/sdktests/common_tests_instance_id.go
@@ -112,8 +112,8 @@ func (c CommonInstanceIDTests) Run(t *ldtest.T) {
 	if c.sdkKind.IsServerSide() {
 		t.Run("polling initializer requests", func(t *ldtest.T) {
 			initializerData := mockld.NewServerSDKDataBuilder().Build()
-			synchronizerData := mockld.NewServerSDKDataBuilder().
-				IntentCode("none").IntentReason("up-to-date").Build()
+			synchronizerData := mockld.FDv2SDKDataFromServerSDKData(
+				mockld.NewServerSDKDataBuilder().Build(), "none", "up-to-date", "initial")
 			dataSystem := NewSDKDataSystem(t, synchronizerData,
 				DataSystemOptionPollingInitializer(initializerData))
 			_ = NewSDKClient(t, c.baseSDKConfigurationPlus(dataSystem)...)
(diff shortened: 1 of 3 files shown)


sdktests/common_tests_instance_id.go
```diff
@@ -60,17 +60,161 @@ func (c CommonInstanceIDTests) Run(t *ldtest.T) {
...
-}
 
-func (c CommonInstanceIDTests) RunPHP(t *ldtest.T) {
-	t.RequireCapability(servicedef.CapabilityInstanceID)
+	// instance-id identifies an SDK client instance; two distinct clients
+	// living in the same process must never share a value, or telemetry can't
+	// disambiguate them. Stand up two independent clients back to back and
+	// assert their instance-ids differ. Gated on !CapabilitySingleton since
+	// the test requires creating a second client while the first still exists.
+	if !t.Capabilities().Has(servicedef.CapabilitySingleton) {
```
tanderson-ld, line 76: Not to be that guy, but an SDK could have the ability to make a singleton and the ability to make instances.  I supposed we don't do that nor do we expect to do that?

sdktests/common_tests_tags.go
```diff
@@ -106,6 +109,104 @@ func (c CommonTagsTests) Run(t *ldtest.T) {
...
+	// sufficient.
+	fdv2TagParams := tagsTestParams{
+		tags: servicedef.SDKConfigTagsParams{
+			ApplicationID:      o.Some("test-app"),
+			ApplicationVersion: o.Some("1.0.0"),
+		},
+	}
+	fdv2TagParams.expectedHeaderValue = c.makeExpectedTagsHeader(fdv2TagParams.tags)
+
+	if !c.isClientSide {
```
tanderson-ld: Is there a way to positively identify the category instead of negatively? Then if another type is added, this doesn't start running accidentally.

tanderson-ld (approved):

sdktests/common_tests_instance_id.go
keelerm84: In that case, the SDK wouldn't specify a capability of singleton like this. It's one of the earliest capabilities, and just designates SDKs that can ONLY create a single instance at a time, not that it has the ability to create a singleton. It should have been CapabilityMultiClient or something instead.

PR #374, opened 2026-07-07
feat: add contract tests for file-based flag overrides

kinyoklion:
Adds a contract test suite for the SDK flag-overrides feature (OVERRIDE spec), gated on a new `flag-overrides` capability so SDKs that don't advertise it skip everything.

Configuration follows the shared-filesystem arrangement already used for the TLS custom CA file: the harness writes override documents to temp files and passes absolute paths in a new `overrides` config block (`filePaths`, `duplicateKeysHandling`, `watch`, `poll`, `pollIntervalMs`). An `OverrideFile` helper creates, rewrites (deliberately non-atomic truncate-and-write, since SDKs must tolerate mid-write reads), and clears the files during a test.

Coverage:
- Static: flagValues precedence with the `OFF` reason and `isOverride: true`; a full flag override evaluating targeting rules (`RULE_MATCH`); passthrough for non-overridden flags with `isOverride` strictly absent; an overridden flag rule resolving an overridden segment through the merged view; all-flags state reflecting overrides.
- Uninitialized client (blocking mock stream + `initCanFail`): the overridden flag serves its override value and reason; a non-overridden flag still returns the default with `CLIENT_NOT_READY`.
- Summary events: an override counter carries `"override": true` (new `overrideFlagCounter` matcher) while an ordinary flag's counter carries no marker (proven by the existing exact-key matcher), and a tracked (`trackEvents: true`) override evaluation produces no individual feature event — the payload's exact item set is index + summary only.
- Multi-file merging with `ignore` duplicate handling (first file wins, later files still merged) and YAML documents.
- Hot reload, parameterized over watch mode and poll mode: add, change, and remove overrides on a running client, plus malformed-file handling (the last good overrides stay in effect across a bad edit, then recovery after the fix).

One note on reason assertions: the typed `ldreason.EvaluationReason` in the response model drops JSON properties it doesn't know, which would silently discard `isOverride`. The suite therefore asserts reasons at the raw-JSON level via a small same-package helper, which also lets the passthrough test enforce the spec's "omitted when false" serialization rule strictly.

Verified end-to-end against the Go SDK's in-progress implementation: the new suite passes in both reload modes, and the full harness run (4945 tests) stays green.

SDK-2657

### DIFF
--- a/sdktests/server_side_events_summary.go
+++ b/sdktests/server_side_events_summary.go
@@ -499,6 +499,19 @@ func flagCounter(value interface{}, variation int, version int, count int) m.Mat
 	)
 }
 
+// overrideFlagCounter is the same as flagCounter, but for a flag whose evaluation was overridden
+// by the flag overrides feature: the counter must additionally carry "override": true so that
+// override evaluations are aggregated separately from ordinary ones.
+func overrideFlagCounter(value interface{}, variation int, version int, count int) m.Matcher {
+	return m.MapOf(
+		m.KV("value", m.JSONEqual(value)),
+		m.KV("variation", m.Equal(variation)),
+		m.KV("version", m.Equal(version)),
+		m.KV("count", m.Equal(count)),
+		m.KV("override", m.Equal(true)),
+	)
+}
+
 func flagCounterWithAnyCount(value interface{}, variation int, version int) m.Matcher {
 	return m.MapOf(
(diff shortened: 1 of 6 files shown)


joker23 (changes requested):
Need to add `flag-overrides` capability docs to `service_spec.md`

joker23 (approved):

PR #377, opened 2026-07-10
feat: Add contract test for streaming connection close on client shutdown

jsonbailey:
## Summary

Adds a new connection-lifecycle subtest to the **legacy server-side streaming suite** asserting that the SDK closes its streaming TCP socket when the client is closed.

- **Where it's registered:** `sdktests/server_side_stream_all.go`, as a `connection lifecycle` subtest under `doServerSideStreamTests` (path: `streaming/connection lifecycle/SDK closes streaming connection when client is closed`).
- **Ungated:** It is NOT behind `CapabilityFDv1Fallback` or any FDv2 capability. It uses a plain streaming config (`NewSDKDataSource(...DataSourceOptionStreaming())` + `NewSDKClient(t, dataSource)`, no data-system object), which exercises the legacy FDv1 streaming data source.
- **Detection mechanism:** Same as the FDv2 close assertion — after `client.Close()`, it polls the captured streaming request's `Context.Done()`. Go's HTTP server cancels the incoming request Context when the client closes the underlying TCP connection. Deadline `time.Second*3`, poll `time.Millisecond*20`.
- **Double-close safe:** `TestServiceEntity.Close()` uses `sync.Once`, so the explicit close plus the automatic end-of-test close do not double-DELETE.

## Local validation

Run against two python-server-sdk versions on PORT 8001:

- **Buggy (`python-server-sdk` main):** test RAN and FAILED **3/3** — `SDK did not close the streaming connection after the client was closed`. Socket stays ESTABLISHED past the 3s deadline (GC-only closure).
- **Fixed (SDK-2668 / python-server-sdk PR #456):** test RAN and PASSED **3/3** — force-close of the urllib3 pool cancels the request Context near-instantly.

No flakiness observed. The CPython refcounting spurious-pass concern did not materialize: buggy main misses the deadline by a wide, reliable margin.

## Needs team discussion before merge

This assertion runs against **every server-side SDK**, not just Python. Any SDK that shares the same un-closed connection-pool pattern on its legacy streaming path will start **failing CI** once this lands. **Ruby is a likely candidate.** This is why the PR is opened as a **draft** — we should confirm which SDKs currently pass/fail and coordinate fixes (or temporary gating) before merging.

## References

- Ticket: **SDK-2669**
- Related fix: **SDK-2668** / python-server-sdk PR #456

---

> [!NOTE]
> **Medium Risk**
> Ungated harness assertions will fail CI for any SDK that does not actively close streaming connections on shutdown; behavior is intentional but broad blast radius across server- and client-side SDKs.
> 
> **Overview**
> Adds **connection lifecycle** contract tests under both **server-side** and **client-side** streaming suites, requiring that `client.Close()` tears down the long-lived SSE connection instead of leaving the TCP socket open.
> 
> After capturing the streaming request via `RequireConnection`, each test calls `Close()` and polls until the mock server request **context is cancelled** (3s deadline, 20ms interval). Server-side uses legacy FDv1 streaming (`DataSourceOptionStreaming()`); client-side uses `CommonStreamingTests.setupDataSources` (poll + stream for JS SDKs). **Roku** skips the short-lived handshake request so the assertion targets the real SSE connection. These checks are **not** gated on FDv2 capabilities, so every SDK on those paths must pass once merged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c502453d0838b99b455eb0bbfedb699db2a8056. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joker23 (approved):

kinyoklion (approved):
I would like it for client-side as well.

Image
I will see if the fix for erlang is straightforward.

sdktests/client_side_stream_all.go
```diff
@@ -67,3 +71,41 @@ func doClientSideStreamRequestTest(t *ldtest.T) {
...
+			t,
+			func() bool {
+				select {
+				case <-streamRequest.Context.Done():
+					return true
+				default:
+					return false
+				}
+			},
+			time.Second*3,
```
kinyoklion, line 114: I am tempted to say we make this 30 seconds. Most SDKs will pass quickly and erlang will be slower.

kinyoklion: I found a better way to fix erlang. So discard.

PR #404, opened 2026-08-07
feat: add RETRY-conformance contract tests for FDv1 streaming and polling

tanderson-ld:
## Summary

Adds contract tests exercising the RETRY specification's non-permanent-stop guarantee for the FDv1 streaming and polling data sources. Introduces two capability gates so SDKs can adopt independently:

- `retry-conformance-fdv1-streaming`
- `retry-conformance-fdv1-polling`

The new tests exercise the SDK's real production 5-minute extended-regime timing — no test-harness bypass, no compression knobs on the servicedef, no test-only surface in the SDK's public API. They are marked as long-running (`t.LongRunning()`) and only run when the harness is invoked with `-enable-long-running-tests`.

Legacy "do not retry after unrecoverable HTTP error" scenarios remain, gated behind the ABSENCE of the streaming capability so SDKs can migrate on their own schedule. Once every SDK reports the new capabilities, both the legacy tests and the gates can be removed.

## What's added

**Streaming (guarded by `retry-conformance-fdv1-streaming` + `-enable-long-running-tests`):**
- retry after unexpected HTTP error on initial connect (401/403/405)
- retry after unexpected HTTP error on reconnect (401/403/405)
- enters extended-regime backoff after unexpected HTTP error (paired-bound timing assertion)
- does not permanently stop under sustained unexpected HTTP errors
- returns to normal-regime backoff after healthy operation (streaming spec's activeSince reset: 75 s silent stream after an extended-regime reconnect, then next reconnect must be at normal-regime timing)

**Polling (guarded by `retry-conformance-fdv1-polling` + `-enable-long-running-tests`):**
- retry after unexpected HTTP error on initial poll (401/403/405)
- returns to normal-regime cadence after two consecutive successful polls (RETRY §1.8.1 count-based reset)

## Design: long-running instead of compressed timings

An earlier revision of this PR exposed extended-regime timing knobs on the servicedef (`ExtendedInitialDelayMS`, `ResetThresholdMS`) so tests could compress the 5-minute extended-initial delay into a ~30-second observable window. Reviewers pushed back because that shape forced SDKs to expose corresponding test-only knobs on their public API surface just to be testable by the harness.

This revision drops all compression knobs and exercises real production timing instead. That makes each retry-conformance subtest take several minutes of wall clock, which is why the tests are marked `t.LongRunning()` and gated behind `-enable-long-running-tests`. Per-SDK CI schedules for the long-running suite are out of scope for this PR (planned follow-up work).

Framework support for `t.LongRunning()` and the `-enable-long-running-tests` flag was ported from v3 in #428, a prerequisite that landed before this PR's refactor.

## Relationship to PR #401

Supersedes #401, which targeted `v3` and exercised FDv2 code paths — out of scope for the initial RETRY-conformance work, which is FDv1-only. This PR targets `v2` and drives the FDv1 streaming/polling paths in the SDK testservices. #401 will be closed with a pointer here.

SDK-2808

## Test plan
- [x] Contract tests pass against go-server-sdk PR #429 (the reference server SDK implementation) at real production 5-minute extended-regime timing (~3-hour run with `-enable-long-running-tests`)
- [x] Legacy "do not retry" tests still pass for SDKs without the new capability (verified by SDK capability check)
- [x] Retry-conformance tests are correctly SKIPPED for SDKs without the capability
- [x] Retry-conformance tests are correctly SKIPPED when `-enable-long-running-tests` is not set


sdktests/server_side_stream_retry.go
```diff
@@ -199,8 +217,139 @@ func doServerSideStreamRetryTests(t *ldtest.T) {
...
+
+		_ = NewSDKClient(t, WithConfig(servicedef.SDKConfigParams{InitCanFail: true}),
+			WithStreamingConfig(retryConformanceStreamConfig(streamEndpoint)))
+
+		// Observe at least 3 connections at extended-regime timing. Each takes ~500 ms; with
+		// doubling, total ~3.5 s in the worst case. Generous per-connection timeout.
+		for i := 0; i < 3; i++ {
+			_ = streamEndpoint.RequireConnection(t, extendedRegimeConnectionTimeout)
+		}
+	})
```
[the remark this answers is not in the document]
tanderson-ld: Legit finding — the sustained-retry test had two compounding weaknesses:

1. `NewSDKClient` used the default 5 s init wait, so extended-regime retry attempts were pre-buffered in the mock endpoint before the assertion loop ran. A broken SDK that retried a few times and permanently stopped could satisfy the drain trivially.
2. The comment described a final "one more still arrives at the end" live check, but the code didn't include it.

Fixed in 436356d:
- `StartWaitTimeMS: 1 ms` so `NewSDKClient` returns quickly and each `RequireConnection` observes a live attempt rather than draining a pre-filled buffer (mirroring the sibling extended-regime-backoff test at lines 285-319).
- Added the missing tail `RequireConnection` after the loop.
- Bumped the loop from 3 to 5 iterations and introduced a per-test 12 s connection timeout that accommodates the extended-regime doubling backoff gaps (500 / 1000 / 2000 / 4000 / 8000 ms), so a broken SDK can't slip through with a handful of early attempts.

sdktests/server_side_poll_retry.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// initialPollTimeout bounds the SDK's initial poll (no backoff involved; happens on client
+	// construction).
+	initialPollTimeout := 15 * time.Second
+
+	// compressedExtendedInitialDelay is passed to the SDK so that the extended-regime base
+	// delay is below the polling wait floor (PollInterval). With this in place the observed
+	// gap between polls will be exactly PollInterval (30 s), the smallest observable window
+	// for these tests.
+	compressedExtendedInitialDelay := ldtime.UnixMillisecondTime(1)
```
keelerm84: I'm a little unclear on this, but it sounds like this is a way to get the SDK to bypass some standard behavior. If the test harness can do it, can't customers? That seems like maybe something we don't necessarily want to expose.

servicedef/service_params.go
```diff
@@ -81,6 +81,31 @@ const (
...
+	// the data source to permanently cease operation. In particular, `401` / `403` / other `4xx`
+	// statuses and TLS/certificate validation failures now trigger an extended-regime backoff
+	// instead of a permanent stop.
+	//
+	// SDKs declaring this capability MUST also honor the extended-regime timing knobs exposed on
+	// SDKConfigStreamingParams (see ExtendedInitialDelayMS, ResetThresholdMS). The capability gates
+	// the new "retry-conformance" subtests. Legacy "do not retry after unexpected HTTP error"
+	// subtests are only run when this capability is absent; both sets of tests can be decommissioned
+	// once every SDK reports the capability.
+	CapabilityRetryConformanceFDv1Streaming = "retry-conformance-fdv1-streaming"
```
kinyoklion, line 94: I think these should be part of the long running test capability, and we probably need to enforce a CI schedule for that capability across all the SDKs.

sdktests/server_side_stream_retry.go
```diff
@@ -199,8 +217,151 @@ func doServerSideStreamRetryTests(t *ldtest.T) {
...
+				streamEndpoint := makeStreamEndpoint(t, handler)
+				t.Defer(streamEndpoint.Close)
+
+				client := NewSDKClient(t, WithStreamingConfig(retryConformanceStreamConfig(streamEndpoint)))
+				result := client.EvaluateAllFlags(t, servicedef.EvaluateAllFlagsParams{Context: o.Some(context)})
+				m.In(t).Assert(result, EvalAllFlagsValueForKeyShouldEqual(flagKey, expectedValueV1))
+
+				for i := 0; i < 3; i++ { // expect three requests
+					_ = streamEndpoint.RequireConnection(t, extendedRegimeConnectionTimeout)
+				}
```
keelerm84: This ordering seems off. Shouldn't we establish that all of the connections have occurred before we try to assert that the evaluations are going to be correct?

sdktests/server_side_stream_retry.go
```diff
@@ -199,8 +217,151 @@ func doServerSideStreamRetryTests(t *ldtest.T) {
...
+				handler := httphelpers.SequentialHandler(
+					stream1.Handler(),                     // 1st: first stream data
+					httphelpers.HandlerWithStatus(status), // 2nd: error
+					httphelpers.HandlerWithStatus(status), // 3rd: error
+					stream2.Handler(),                     // 4th: second stream data
+				)
+				streamEndpoint := makeStreamEndpoint(t, handler)
+				t.Defer(streamEndpoint.Close)
+
+				client := NewSDKClient(t, WithStreamingConfig(retryConformanceStreamConfig(streamEndpoint)))
```
keelerm84: Same question here about the connections vs evaluations.

joker23 (commented):
we should update https://github.com/launchdarkly/sdk-test-harness/blob/v3/docs/service_spec.md as well to document this capability.

sdktests/server_side_poll_retry.go
tanderson-ld: Yeah, I took this angle to avoid this being a long running test.  We already have precedent for initial delay being modifiable (even below the 1 second min), so this felt tangentially related.  In the go-server-sdk, I put this in an Internal() sub-builder to discourage usage.

https://github.com/launchdarkly/go-server-sdk/pull/429/changes#diff-8b7845d1c8815d6788e19476623ddbc18cb48068970c1093bdf4cc12361c24c5R23

Thoughts?

servicedef/service_params.go
tanderson-ld: Related discussion: https://github.com/launchdarkly/sdk-test-harness/pull/404#discussion_r3770093528 would like to get your input on the tradeoffs of being long running vs short running with Internal() API

sdktests/server_side_stream_retry.go
tanderson-ld: Good call. Reordered in 61d8739 — the 3-connection drain now runs before `EvaluateAllFlags`, so the test reads chronologically (prove the retry happened, then use the evaluation as the payoff).

sdktests/server_side_stream_retry.go
tanderson-ld: Same fix in 61d8739. Slightly different shape here: the initial `RequireConnection` (conn #1) now runs before `EvaluateAllFlags` to confirm the connection landed, then `request1.Cancel()` is grouped with the drain of connections #2–4 so the retry-then-success sequence reads in order.

docs/service_spec.md
```diff
@@ -134,6 +134,20 @@ docker run -p 6379:6379 redis
 
 This means the SDK is requesting gzip compression support on polling payloads. The SDK is expected to set the `Accept-Encoding` header to `gzip` in addition to enabling this capability.
 
+#### Capability `"retry-conformance-fdv1-streaming"`
+
+This means that the SDK's FDv1 streaming data source conforms to the RETRY specification: no HTTP response and no transport-level failure causes the data source to permanently cease operation. In particular, `401` / `403` / other `4xx` statuses and TLS/certificate validation failures trigger an extended-regime backoff (retry with a longer delay) instead of a permanent stop.
+
+SDKs declaring this capability MUST also honor the extended-regime timing knobs `extendedInitialDelayMs` and `resetThresholdMs` on the `streaming` configuration object; the test harness sets these to compress the extended regime into an observable window. See `streaming` configuration below.
+
+The capability gates the new retry-conformance subtests. Legacy "do not retry after unexpected HTTP error" subtests are only run when this capability is absent; both sets of tests can be decommissioned once every SDK reports the capability.
```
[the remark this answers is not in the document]
tanderson-ld: Legit finding — the client-side legacy "do not retry" tests were gated only by `client-event-source-http-errors` and would fail against a client-side SDK that declared `retry-conformance-fdv1-streaming` (since the RETRY spec has that SDK retry rather than permanently stop).

Fixed in dc86e6b: added the same skip-if-capability-present guard the server-side legacy tests use. No client SDK currently declares the new capability, so the current test outcomes are unchanged; the guard is future-proofing for when one does.

sdktests/server_side_poll_retry.go
kinyoklion: I prefer not having a bypass in our API as well.

I would be curious where we allow sub-second on purpose versus being a bug.

sdktests/client_side_stream_retry.go
```diff
@@ -189,6 +189,14 @@ func doClientSideStreamRetryTests(t *ldtest.T) {
...
 
 	t.Run("do not retry after unrecoverable HTTP error on initial connect", func(t *ldtest.T) {
+		// Legacy "do not retry" behavior for SDKs that have not adopted the RETRY
+		// specification. SDKs declaring retry-conformance-fdv1-streaming retry after
+		// unexpected HTTP errors per the spec; skip the legacy test for them.
+		if t.Capabilities().Has(servicedef.CapabilityRetryConformanceFDv1Streaming) {
+			t.SkipWithReason("SDK reports " + servicedef.CapabilityRetryConformanceFDv1Streaming +
+				"; legacy permanent-stop behavior does not apply")
+			return
+		}
```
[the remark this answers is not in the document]
tanderson-ld: This will be done when we get to the client side part of the epic.

sdktests/server_side_poll_retry.go
tanderson-ld: https://github.com/launchdarkly/go-server-sdk/blob/v7/ldcomponents/streaming_data_source_builder.go#L35

https://github.com/launchdarkly/java-core/blob/main/lib/sdk/server/src/main/java/com/launchdarkly/sdk/server/integrations/StreamingDataSourceBuilder.java#L18

https://github.com/launchdarkly/dotnet-core/blob/main/pkgs/sdk/server/src/Components.cs#L420

Here we use 1ms
https://github.com/launchdarkly/sdk-test-harness/blob/v2/sdktests/server_side_stream_retry.go#L27

This test would not meet the 2 second timeout requirement if the 3rd request were exponentially backed off from 1 second.
https://github.com/launchdarkly/sdk-test-harness/blob/v2/sdktests/server_side_stream_retry.go#L184

kinyoklion: Yeah, I am against this as well. And have declined similar changes to other SDKs qhen it was requested.

kinyoklion: The allowing below minimum part. Configurability whatever.

kinyoklion: The same initial reconnect delay is constrained to at least 1 second in many SDKs even though the contract tests attempt to set it lower.

tanderson-ld: The bypass is gone as of 9964faf. `ExtendedInitialDelayMS` (streaming and polling) and `ResetThresholdMS` are removed from the servicedef; the go-server-sdk PR #429 will be refactored to drop the matching `Internal()` accessors and exported knobs.

The retry-conformance tests now run at the SDK's production 5-minute extended-regime delay, gated by `t.LongRunning()` and `-enable-long-running-tests` (thanks to Jason's long-running framework support ported in #428).

servicedef/service_params.go
tanderson-ld: Done in 9964faf. Both retry-conformance capabilities now call `t.LongRunning()` — the tests run at real production timing (5-min extended-regime delay) and only execute under `-enable-long-running-tests`.

Per-SDK CI schedules for the long-running suite are being deferred out of this epic. That's a cross-SDK infra change that would apply beyond retry-conformance; I'll file a follow-up ticket once we're through the SDK fan-out.

sdktests/server_side_stream_retry.go
```diff
@@ -199,8 +199,195 @@ func doServerSideStreamRetryTests(t *ldtest.T) {
...
+
+				// 1st connection produces the error
+				_ = streamEndpoint.RequireConnection(t, incomingConnectionTimeout)
+
+				// 2nd connection arrives after an extended-regime wait, proving the SDK retried
+				// rather than permanently stopping.
+				_ = streamEndpoint.RequireConnection(t, extendedRegimeConnectionTimeout)
+
+				result := client.EvaluateAllFlags(t, servicedef.EvaluateAllFlagsParams{Context: o.Some(context)})
+				m.In(t).Assert(result, EvalAllFlagsValueForKeyShouldEqual(flagKey, expectedValueV1))
```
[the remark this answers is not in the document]
tanderson-ld: Legit finding — fixed in b524b32. Switched to `pollUntilFlagValueUpdated(...)` so the assertion waits for the SDK to actually populate its store after `RequireConnection` returns, matching the sibling reconnect test and the polling tests.

sdktests/server_side_poll_retry.go
```diff
@@ -0,0 +1,133 @@
...
+		_ = pollEndpoint.RequireConnection(t, normalRegimePollTimeout)
+
+		// 4th poll: fault at normal-regime timing (SDK is now back in normal regime after
+		// two consecutive successes, so this poll happens ~30 s after poll 3).
+		_ = pollEndpoint.RequireConnection(t, normalRegimePollTimeout)
+
+		// 5th poll: normal-regime retry after the 4th poll's 401. Should arrive within
+		// PollInterval (30 s), NOT extended-regime timing (~5 min). If the regime hadn't
+		// reset, this timeout would fail and the test surfaces the bug.
+		_ = pollEndpoint.RequireConnection(t, normalRegimePollTimeout)
```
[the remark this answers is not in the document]
tanderson-ld: Legit finding — fixed in b524b32. The 4th-poll 401 would put a spec-conformant SDK back into the extended regime, and the assertion that poll 5 arrive within 60 s would then fail.

Restructured the test: dropped the second 401 and the 5th poll. The new shape is `401 → success → success → success`. Poll 4 arriving within normal `PollInterval` cadence (60 s budget) is the reset proof — an SDK that stayed in the extended regime after §1.8.1's two-consecutive-success reset would take ~5 min and fail the assertion.

kinyoklion (approved):

